PR #49, opened 2022-07-27
feat(pipelines): build tagging

c3ppc3pp:
This PR adds tagging capabilities to the pipeline:
when a git tag is set and the build is running against that tag, it will create a docker image with the same tag.

Global changes:
- pipeline_set_output_tag.sh now will output version too, which is 0.0.0 by default, will be set to the tag if available. Docker Image tag will be the commit checksum, unless its built against develop or main branches, then the image tag will be latest and stable respectively.

With crux/crux-ui pipeline:
- Check if package.json version is the same as the tag, if not, build will fail.
- Docker Images are built from the dockerfiles, we wont use any artifact with these.
- Docker will tag additionally if necessary(we can apply multiple tags automatically with a bit of work from now on), then push the images. Same thing goes for signing.

crux/crux-ui build:
- Timezone now can be set during runtime instead of buildtime via TZ environment variable.
- Removed npx from crux build processes
- Fixed some missing deps in the crux build processes


c3ppc3pp:
@m8vago Please test the images produced by the new build process, especially the recaptcha things! Thank you! 

CRUX-UI https://github.com/dyrector-io/dyrectorio/pkgs/container/dyrectorio%2Fweb%2Fcrux-ui/29902123?tag=0.0.1

CRUX https://github.com/dyrector-io/dyrectorio/pkgs/container/dyrectorio%2Fweb%2Fcrux/29902106?tag=0.0.1

web/crux-ui/Dockerfile
```diff
@@ -1,29 +1,39 @@
...
+COPY --from=BUILDER /app/package.json ./package.json
+COPY --from=BUILDER --chown=node:node /app/.next ./.next
+COPY --from=BUILDER /app/node_modules ./node_modules
 
 USER node
 
 EXPOSE 3000
+ENV PORT 3000
 
 ENTRYPOINT [ "npm", "run", "prod"]
```
polaroi8d, line 39: Could you add NL at EOF, please?

.github/workflows/web-crux-ui.yml
```diff
@@ -73,18 +62,20 @@ jobs:
...
     needs: check
     container: docker:latest
-    if: github.ref_name == 'develop' || github.ref_name == 'main'
+    if: github.ref_name == 'develop' || github.ref_name == 'main' || github.ref_type == 'tag'
     defaults:
       run:
         working-directory: ${{ env.WORKING_DIRECTORY }}
     steps:
 
-    - name: Download artifact
```
polaroi8d, line 82: We are in the same 'context', this is the reason why we don't need to download the artifacts?

polaroi8d (changes requested):

.github/workflows/web-crux-ui.yml
c3ppc3pp: Dockerfile does a prod build too, so downloading the artifacts are not necessary.

web/crux/package.json
```diff
@@ -33,7 +33,8 @@
     "database:up": "docker-compose -f docker-compose.database.yaml up -d",
     "database:down": "docker-compose -f docker-compose.database.yaml down",
     "test:crux": "npm run docker:up && npx prisma migrate deploy && jest -i && npm docker:down",
-    "compodoc": "npx compodoc -p tsconfig.doc.json --theme stripe"
+    "prismagen": "prisma generate",
+    "compodoc": "compodoc -p tsconfig.doc.json --theme stripe"
```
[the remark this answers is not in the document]
c3ppc3pp: The "compodoc" command was there already, I assumed we are using it. If it is unused we can remove it no issues.

web/crux/crux-docker-entrypoint.sh
```diff
@@ -13,6 +13,9 @@ generate_cert() {
     fi
 }
 
+# ln -sf /usr/share/zoneinfo/${TIME_ZONE} /etc/localtime
```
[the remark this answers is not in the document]
c3ppc3pp: Good catch! Fixed.

web/crux-ui/Dockerfile
c3ppc3pp: Yes, done.

polaroi8d (approved):
LGTM, thanks.

.github/workflows/web-crux.yml
```diff
@@ -76,19 +70,19 @@ jobs:
...
     steps:
     - name: Checkout
       uses: actions/checkout@v3
 
-    - name: Download /dist artifact
-      uses: actions/download-artifact@v3
-      with:
-        name: crux-web-dist
-        path: ${{ env.WORKING_DIRECTORY }}/dist
+      # if tag isn't the version setted in package.json will make the job fail
```
robot9706, line 81: `set` instead of `setted` ?

c3ppc3pp: You're right! Thanks! Also fixed in another pipeline where I used this comment.

robot9706 (approved):
😎

c3ppc3pp:
Hold it off for a moment. There's a showstopper bug in cert-generation script.

c3ppc3pp:
Re-review it please!

.github/workflows/web-crux.yml
```diff
@@ -97,12 +91,21 @@ jobs:
...
 
-    - name: Docker build and push
+    - name: Docker build 
       run: |
-        docker-compose build
-        docker-compose push crux
+        docker build . -t ${REGISTRY}/${IMAGE_NAME}:${{ needs.prerequirements.outputs.tag }}
       env:
-        CRUX_IMAGE_TAG: ${{ needs.prerequirements.outputs.tag }}
+        CRUX_UI_IMAGE_TAG: ${{ needs.prerequirements.outputs.tag }}
```
m8vago, line 98: ui?

m8vago: they have a shared image tag now?

PR #48, opened 2022-07-27
refactor(agent): eliminated globals & lint fixes

robot9706:
eliminiated global config instances
fixed some lint errors
added "gochecknoglobals" to lint config
added gin middleware for controllers to access the appconfig
created a custom grpc context for worker functions to access the appconfig

PR #53, opened 2022-07-28
refactor: refactoring panic calls

robot9706:
depends on: #48


agent/pkg/dagent/dagent.go
```diff
@@ -94,7 +95,7 @@ func Serve(cfg *config.Configuration) {
 		err := utils.ExecTraefik(context.TODO(), params, cfg)
 		if err != nil {
 			// we wanted to start traefik, but something is not ok, thus panic!
```
polaroi8d, line 97: The log is communicative enough, I think we can remove this comment.

agent/pkg/dagent/utils/release.go
```diff
@@ -48,14 +49,14 @@ func DraftRelease(instance string, versionData v1.VersionData, deployResponse v1
 
 	if err != nil {
 		log.Println(fmt.Sprintln("Version drafting error ", err))
+		return
 	}
```
polaroi8d, line 53:
```
content, err := yaml.Marshal(&release); err != nil {
		log.Println(fmt.Sprintln("Version drafting error ", err))
		return
	}
```

Refactor like this, what do you think? and @nandor-magyar

polaroi8d (changes requested):

agent/pkg/dagent/utils/release.go
robot9706: This makes it impossible to access `content` outside of the if scope.

polaroi8d:
Minor comment from my side, btw LGTM (informal). Please @nandor-magyar check it too.

polaroi8d (approved):

nandor-magyar (approved):
Neat :sunglasses:

PR #54, opened 2022-07-29
refactor(agent): tests

robot9706:


polaroi8d (approved):

agent/internal/dogger/dogger_test.go
```diff
@@ -0,0 +1,170 @@
...
+			assert.Equal(t, expected, message.Log)
+			return nil
+		},
+	}
+	dogger := dogger.NewDeploymentLogger(&deploymentId, ts, context.TODO(), cfg)
+
+	dogger.Write("hello", "abcd")
+}
+
+func TestDoggerErrorTest(t *testing.T) {
```
nandor-magyar, line 93: No assertion in test case.

agent/internal/dogger/dogger_test.go
```diff
@@ -0,0 +1,170 @@
...
+}
+
+func TestDoggerDeploymentStatusError(t *testing.T) {
+	cfg := &config.CommonConfiguration{}
+	deploymentId := "abcd"
+
+	ts := AgentTestStream{
+		testing: t,
+		AssertLog: func(t *testing.T, message *crux.DeploymentStatusMessage) error {
+			return errors.New("Test error")
```
nandor-magyar, line 132: missing assertion

nandor-magyar (changes requested):
In general, i think the test cases should have this structure:
1. prepare
2. act - use the functions at matter
3. assert - validate results
4. cleanup, if need be

The type is nice that you have created for testing, The AssertFn could return a `func(expected string)` error, that is executed at the end of the the test function.
A good assert communicates the behavior and the expectations clearly.

nandor-magyar (approved):

PR #55, opened 2022-07-29
fix(agent): swagger double init during testing

robot9706:


robot9706:
Having problems with the swagger package. Closing for now.

nandor-magyar (approved):

polaroi8d:
In general, we are using our service names in lower case, if there is no extra reason please use it like that.

robot9706:
> In general, we are using our service names in lower case, if there is no extra reason please use it like that.

I'm guessing you are talking about the swagger command. I'm using an upper case name because `SwaggerInfo...` variable names are generated using the instance name, if I'd use a lower case name it'd look like `SwaggerInfocrane` not `SwaggerInfoCrane` which I think looks off.

polaroi8d:
> > In general, we are using our service names in lower case, if there is no extra reason please use it like that.
> 
> I'm guessing you are talking about the swagger command. I'm using an upper case name because `SwaggerInfo...` variable names are generated using the instance name, if I'd use a lower case name it'd look like `SwaggerInfocrane` not `SwaggerInfoCrane` which I think looks off.

I am talking about instance names. (sorry for not mentioned)

polaroi8d (approved):
LGTM

PR #61, opened 2022-08-03
refactor(agent): gRPC communication with core

robot9706:
Implement *core* functionality to the gRPC communication to start phasing out the old REST interface.


agent/internal/grpc/grpc.go
```diff
@@ -268,7 +276,12 @@ func executeVersionDeployRequest(
 		imageReq := mapper.MapDeployImage(req.Requests[i], appConfig)
 		dog.SetRequestID(imageReq.RequestID)
 
-		if err = deploy(ctx, dog, imageReq, &v1.VersionData{Version: req.VersionName, ReleaseNotes: req.ReleaseNotes}); err != nil {
+		var versionData *v1.VersionData = nil
```
nandor-magyar, line 279: default is nil for pointers, it is unnecessary to explicitly declare.

nandor-magyar:
Please remove the unnecessary masrhal / unmarshal test functions.

protobuf/proto/agent.proto
```diff
@@ -41,6 +41,8 @@ message AgentCommand {
   oneof command {
     VersionDeployRequest deploy = 1;
     ContainerStatusRequest containerStatus = 2;
+    ContainerDeleteRequest containerDelete = 3;
+    DeployRequestCore deployCore = 4;
```
nandor-magyar, line 45: Use of the word Legacy instead of Core? What do you think? Core might not say anything to the reader.

nandor-magyar (approved):

m8vago (changes requested):
prename to prefix

m8vago (approved):

PR #64, opened 2022-08-04
refactor: websocket message routing

m8vago:
Fixing an edge case, when one user with a single browser process, thus a single ws connection, editing two different version or deployment.


robot9706 (approved):
😎

web/crux-ui/locales/en/registries.json
```diff
@@ -29,8 +29,8 @@
 
   "type": {
     "v2": "V2 Registry",
-    "hub": "Docker HUB",
-    "gitlab": "Gitlab Registry",
+    "hub": "Docker Hub",
+    "gitlab": "GitLab Registry",
```
nandor-magyar, line 33: For the sake of correctness, the real name of that software is **GitLab Container Registry**. Shouldn't we use their _official_ name?

m8vago: I guess it's too long @polaroi8d ?

nandor-magyar (approved):

PR #66, opened 2022-08-08
refactor(agent): moved crane/dagent errors to v1

robot9706:
Crane and Dagent had almost the same errors defined for the REST interface. Moved these to the v1 common package.


agent/pkg/api/v1/error.go
```diff
@@ -1,4 +1,4 @@
-package model
```
polaroi8d, line 1: Is this good place for errors in API pkg @nandor-magyar ?

robot9706: These errors are used by the REST interface, that's why I moved them to the v1 model. 🤔

nandor-magyar: Yes, I like it, there is no better place for it right now.

PR #65, opened 2022-08-05
refactor(agent): move container builder to package

robot9706:
Moved container builder to a separate package so it could be used independetly from agent packages.
Also ran `gofmt`.

Depends on: #61

PR #68, opened 2022-08-08
refactor(agent): context usage

robot9706:
Refactor contexts and TODO context usage.
Depends on: #65 

Tests should use `Background` contexts. Pass contexts to function calls and use if possible.


agent/pkg/crane/k8s/configmap.go
```diff
@@ -51,7 +51,7 @@ func (cm *configmap) deployConfigMapData(namespace, name string, envList map[str
 		return err
 	}
 
-	result, err := client.Apply(context.TODO(),
+	result, err := client.Apply(context.Background(),
```
nandor-magyar, line 54: We could use the configmap.ctx context. Here and other facade-like structs too.

robot9706: True, I didn't notice that field 👀

nandor-magyar (approved):

c3ppc3pp (approved):
One letter variables are rustling my feathers, but the usage seem justified; otherwise looks fine.

PR #71, opened 2022-08-09
refactor: make state and status name usage consistent

robot9706:
Container `state` and `status` are mixed up in the code. Make it consistent to how docker uses them.


protobuf/proto/crux.proto
```diff
@@ -629,8 +629,8 @@ message ContainerStatusItem {
   string name = 101;
   string command = 103;
   google.protobuf.Timestamp createdAt = 104;
-  ContainerStatus status = 105; // The 'State' of the container (Created, Running, etc)
-  string state = 106; // The 'Status' of the container ("Created 1min ago", "Exited with code 123", etc)
+  ContainerState state = 105; // The 'State' of the container (Created, Running, etc)
```
nandor-magyar, line 632: Add comment above the item, there is a proto based docs generator, that way it will be moved into the generated docs too surely, but if comment follows the field I am not so sure.

web/crux/src/grpc/protobuf/proto/crux.ts
```diff
@@ -949,8 +949,10 @@ export interface ContainerStatusItem {
   name: string
   command: string
   createdAt: Timestamp | undefined
-  status: ContainerStatus
-  state: string
+  /** The 'State' of the container (Created, Running, etc) */
```
nandor-magyar, line 952: It is copied here :thinking:

nandor-magyar (commented):
Nice overall! :cake:

protobuf/proto/crux.proto
```diff
@@ -629,8 +629,10 @@ message ContainerStatusItem {
   string name = 101;
   string command = 103;
   google.protobuf.Timestamp createdAt = 104;
-  ContainerStatus status = 105; // The 'State' of the container (Created, Running, etc)
-  string state = 106; // The 'Status' of the container ("Created 1min ago", "Exited with code 123", etc)
+  /** The 'State' of the container (Created, Running, etc) */
+  ContainerState state = 105;
```
polaroi8d, line 633: In here we need both? state and status too?

nandor-magyar (approved):

web/crux/proto/crux.proto
```diff
@@ -629,8 +629,10 @@ message ContainerStatusItem {
   string name = 101;
   string command = 103;
   google.protobuf.Timestamp createdAt = 104;
-  ContainerStatus status = 105;
-  string state = 106;
+  /** The 'State' of the container (Created, Running, etc) */
+  ContainerState state = 105;
+  /** The 'Status' of the container ("Created 1min ago", "Exited with code 123", etc) */
+  string status = 106;
```
polaroi8d, line 635: Same here.

robot9706: These are copies from the `protobuf` folder.

protobuf/proto/crux.proto
robot9706: Crux might not need it, but core does, so as long as we'd like to support legacy we need both.

polaroi8d: Okey, we talked about this in person, can we show the developers somehow that we need both right now? I think we can write a small comment with a legacy tag.

balanceee (approved):
LGTM

PR #76, opened 2022-08-11
refactor(agent): handle k8s resource parsing errors

robot9706:
If user supplied resource limits and requests are incorrect the agent handles the error and fails the deployment (instead of panicing).


polaroi8d (approved):
LGTM, good job with the tests 👍

agent/pkg/crane/k8s/deployment_test.go
```diff
@@ -0,0 +1,270 @@
...
+		},
+		Requests: v1.Resources{
+			CPU:    "3m",
+			Memory: "4m",
+		},
+	}
+
+	_, err := k8s.GetResourceManagementForTest(resourceConfig, nil)
+
+	assert.Error(t, err)
```
nandor-magyar, line 54: This is general, better would to be to assert the error itself, but here are no error types nor constants defined as the error texts, this also means. That it would be nice to have an error type for resource errors.

nandor-magyar (commented):
New error types desire existence. :egg:

agent/pkg/crane/k8s/deployment_test.go
robot9706: 😊

nandor-magyar (approved):
Gleaming error type! :gem:

PR #84, opened 2022-08-16
refactor(crux): url prefix to image name prefix

peterambrus3:
refactor(crux): url prefix to image name prefix
chore: run prettier format


web/crux/prisma/migrations/20220815104826_url_prefix_to_image_name_prefix/migration.sql
```diff
@@ -0,0 +1,18 @@
+-- Copy data
+DROP TABLE IF EXISTS public._migration_instance;
+SELECT 
+	id,
+  "Registry"."urlPrefix"
+INTO public._migration_instance
```
robot9706, line 6: `_migration_registry` would be a better name

balanceee (approved):
LGTM

robot9706 (changes requested):

web/crux/prisma/migrations/20220815104826_url_prefix_to_image_name_prefix/migration.sql
peterambrus3: true, my bad sry

web/crux-ui/src/const.ts
```diff
@@ -25,4 +25,3 @@ export const ATTRIB_CSRF = 'csrf_token'
 export const HEADER_SET_COOKIE = 'set-cookie'
 export const AUTH_RESEND_DELAY = 30 // seconds
 export const WEBOOK_TEST_DELAY = 500 //millis
-
```
polaroi8d, line 28: If we remove there will be no NW at EOF, not?

polaroi8d (approved):
LGTM

robot9706 (approved):
😎

PR #107, opened 2022-08-23
feat: replace nginx with traefik

m8vago:
Using traefik instead of custom nginx for development


web/.env.example
```diff
@@ -10,13 +10,9 @@ SMTP_USER=test
 SMTP_PASSWORD=test
 SMTP_URL=mailslurper:1025/?skip_ssl_verify=true&legacy_ssl=true
 
+COMPOSE_FILE=docker-compose.dev.yaml
```
polaroi8d, line 13: There is any reason why you moved from bottom of the file?

polaroi8d (approved):
We are using a kebab-case naming convention for branches, otherwise good job. LGTM

robot9706 (approved):
🙌

PR #114, opened 2022-08-25
fix(crux-ui): fixed deployment editor

robot9706:
- Navigates back to details view after editing
- Fixed unable to save description field


polaroi8d (approved):
LGTM

web/crux-ui/src/models/grpc/protobuf/proto/crux.ts
```diff
@@ -7276,8 +7276,8 @@ export const UpdateDeploymentRequest = {
     if (message.name !== '') {
       writer.uint32(802).string(message.name)
     }
-    if (message.descripion !== undefined) {
-      writer.uint32(810).string(message.descripion)
+    if (message.description !== undefined) {
```
balanceee, line 7279: If(message.description) would be more safer don't you think?

balanceee (approved):

PR #116, opened 2022-08-25
fix(crux-ui): node setup script improvement

peterambrus3:
implement sh editor and minor ui fixes

chore(crux-ui): run prettier


robot9706 (approved):
😎

PR #109, opened 2022-08-24
feat: extend container config

balanceee:
Agent was already supporting plenty of configuration options, meanwhile the crux only supported a small piece.
The purpose of this commit is to bring up the crux and the agent to the same level in terms of container configuration.

PR #117, opened 2022-08-25
feat(agent): remove unused code

robot9706:
Agent now communicates on gRPC only:
- Removed HTTP server
- Removed SignalR
- Removed some unused code as well

Depends on #109


polaroi8d:
I will start the review after #109 merged 🐐

polaroi8d (approved):
LGTM

agent/Makefile
```diff
@@ -38,10 +38,22 @@ compile-crane: ##
...
+
 .PHONY: compile-dagent
 compile-dagent: ##
 	cd cmd/dagent && \
-	GOOS=linux GOARCH=amd64 CGO_ENABLED=0 go build ${LDFLAGS} -o ../../build/out/dagent; \
+	GOOS=windows GOARCH=amd64 CGO_ENABLED=0 go build ${LDFLAGS} -o ../../build/out/dagent.exe; \
+	cd -
+
+.PHONY: compile-dagent-windows
+compile-dagent-windows: ##
```
nandor-magyar, line 54: Instead of duplication manipulate variables instead, same could be done as with image_version, defaults to linux, but on machines you can do `goos=windows make compile-dagent`.

agent/pkg/crane/k8s/status.go
```diff
@@ -9,16 +9,22 @@ import (
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 
 	"github.com/dyrector-io/dyrectorio/agent/internal/util"
-	v1 "github.com/dyrector-io/dyrectorio/agent/pkg/api/v1"
 	"github.com/dyrector-io/dyrectorio/agent/pkg/crane/config"
 )
 
+type ContainerStatusResponse struct {
```
nandor-magyar, line 15: Shouldn't this be in api.v1?

nandor-magyar (commented):
Neat :lab_coat:

agent/pkg/crane/k8s/status.go
robot9706: This file is the only place where `ContainerStatusResponse` is used, also the `DeploymentStatus` function is unused, but I left it there for future possible use.

nandor-magyar: We have `crux.GetDeployments` for for status queries, removing this is OK.

nandor-magyar (approved):

polaroi8d (approved):

PR #129, opened 2022-08-31
feat(web): npm workspaces & shared code

robot9706:
This PR makes use of NPM workspaces to arrange crux and crux-ui packages differently. This makes it possible to create a `common` package to shared code between them.


web/crux/src/main.ts
```diff
@@ -64,9 +64,9 @@ const bootstrap = async () => {
     options: {
       package: ['agent'],
       protoPath: [
-        join(__dirname, '../proto/crux.proto'),
-        join(__dirname, '../proto/agent.proto'),
-        join(__dirname, '../proto/common.proto'),
+        join(__dirname, '../../../proto/crux.proto'),
```
robot9706, line 67: Might need to find a better way to resolve proto paths 🤔

robot9706:
Talked with @m8vago, we are not happy with the solution.. trying other things instead.

PR #135, opened 2022-09-02
fix: demo seeder config

peterambrus3:


agent/pkg/builder/container/container_builder.go
```diff
@@ -400,41 +402,54 @@ func prepareImage(dc *DockerContainerBuilder) error {
...
 }
 
-func createNetwork(dc *DockerContainerBuilder) *string {
-	networkName := fmt.Sprintf("%s-network", dc.networkMode)
+func createNetworks(dc *DockerContainerBuilder) []string {
+	var networkIds []string
 
-	filter := filters.NewArgs()
-	filter.Add("name", networkName)
+	for _, network := range dc.networks {
```
robot9706, line 408: The `network` variable could be called `networkName` so the extra variable could be eliminated.

agent/pkg/builder/container/container_builder.go
```diff
@@ -400,41 +402,54 @@ func prepareImage(dc *DockerContainerBuilder) error {
...
+		}
+
+		networkResult, err := dc.client.NetworkCreate(dc.ctx, networkName, networkOpts)
+		if err != nil {
+			logWrite(dc, fmt.Sprintln("Failed to create network: ", err.Error()))
+			return nil
+		}
+
+		networkID := networkResult.ID
+		networkIds = append(networkIds, networkID)
```
robot9706, line 435: Could be simplyfied something like `networkIds = append(networkIds, networkResult.ID)`

.gitignore
```diff
@@ -1,5 +1,6 @@
 .env
 node_modules/
+*.exe
```
robot9706, line 3: If this is used to ignore the build output of the agent then I think instead we could use a .gitingore in the `agent` folder and just ignore the `build/out` path.

robot9706 (changes requested):

.gitignore
```diff
@@ -1,5 +1,6 @@
 .env
 node_modules/
+build/out
```
robot9706, line 3: This might need some investigation, since the gitignore in the agent folder already seem to have this.

robot9706 (approved):

.gitignore
peterambrus3: ye, it's not needed, deleted

agent/pkg/builder/container/container_builder.go
```diff
@@ -400,41 +402,51 @@ func prepareImage(dc *DockerContainerBuilder) error {
 	return nil
 }
 
-func createNetwork(dc *DockerContainerBuilder) *string {
-	networkName := fmt.Sprintf("%s-network", dc.networkMode)
+func createNetworks(dc *DockerContainerBuilder) []string {
+	var networkIds []string
```
m8vago, line 406: networkIDs ?

peterambrus3: would you like to rename the variable or ?

m8vago (approved):

PR #138, opened 2022-09-05
feat: updated ts-proto

robot9706:
Updated ts-proto to the latest version (with the dyrectorio fix) and built proto files.

Make sure to update the proto container:
- Run `npm i` in crux and crux-ui
- Delete the old proto container
- Run `make build-proto-image` in the root directory


balanceee (approved):
LGTM

PR #146, opened 2022-09-07
fix(crux-ui): node edit after save

balanceee:
After saving a new node you can edit it immediately and the node list stays up to date. You only have to regenerate your script if the node's name changed.

### DIFF
--- a/web/crux-ui/src/components/nodes/edit-node-card.tsx
+++ b/web/crux-ui/src/components/nodes/edit-node-card.tsx
@@ -140,8 +140,25 @@ const EditNodeCard = (props: EditNodeCardProps) => {
         } else {
           result = {
             ...values,
+            id: node.id,
             status: node.status,
+            type: node.type,
           } as DyoNodeDetails
+
+          if (values.name !== node.name) {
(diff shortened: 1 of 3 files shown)


peterambrus3 (changes requested):
In edit mode generate Kubernetes script edit name press save, generate again Kubernetes script edit desc press save -> Requirements section changed

After the first save the URL is not changed and this cause these problems: 
-discard not working
-breadcrumb home btn is not working
-nodes side menu does not navigate to the node list screen

web/crux-ui/src/components/nodes/edit-node-card.tsx
```diff
@@ -139,13 +139,16 @@ const EditNodeCard = (props: EditNodeCardProps) => {
           } as DyoNodeDetails
         } else {
           result = {
+            id: node.id,
```
robot9706, line 142: I think the new state is not required, and this line should fix the problem.

robot9706 (changes requested):

polaroi8d (approved):
LGTM (not tested locally)

robot9706 (approved):
😎

peterambrus3 (approved):
lgtm

PR #147, opened 2022-09-08
feat(web): audit log show email

robot9706:
On the audit log page show the email of the user instead of the name.


peterambrus3 (approved):
lgtm

PR #152, opened 2022-09-09
fix(agent): use regex on container listing

balanceee:
The previous version checked whether the container name contains the prefix or not. It lead to see containers which is contained the prefix but not necessary started with it.


robot9706 (approved):

peterambrus3 (approved):
lgtm

PR #153, opened 2022-09-09
feat: secrets

robot9706:
Adds secret functionality. Secret keys can be defined on images and they can be set before deployment. Secrets are encrypted using PGP.


.vscode/settings.json
```diff
@@ -1,3 +1,14 @@
 {
-    "go.buildFlags": ["-tags=unit,integration"],
+    "go.buildFlags": [
+        "-tags=unit,integration"
+    ],
+    "files.exclude": {
+        "**/.git": true,
+        "**/.svn": true,
+        "**/.hg": true,
```
polaroi8d, line 8: Why we need to exclude svg and mercurial file formats?

agent/internal/config/secret.go
```diff
@@ -0,0 +1,98 @@
...
+	"log"
+	"os"
+
+	"github.com/ProtonMail/gopenpgp/v2/crypto"
+)
+
+type ConfigFromFile string
+
+func (field *ConfigFromFile) SetValue(location string) error {
+	log.Println("Custom key parsing config things")
```
polaroi8d, line 18: `things`?

agent/internal/config/secret.go
```diff
@@ -0,0 +1,98 @@
...
+		log.Printf("Key file is expired: %v\n", location)
+		return generateKey(location)
+	}
+}
+
+func generateKey(location string) (string, error) {
+	log.Printf("Generating new key file...")
+	const (
+		name  = "dyrectorio crane"
+		email = "agent@dyrector.io"
```
polaroi8d, line 61: We don't have an email like this, right now.

agent/internal/config/secret.go
```diff
@@ -0,0 +1,98 @@
...
+	} else {
+		log.Printf("Key file is expired: %v\n", location)
+		return generateKey(location)
+	}
+}
+
+func generateKey(location string) (string, error) {
+	log.Printf("Generating new key file...")
+	const (
+		name  = "dyrectorio crane"
```
polaroi8d, line 60: dyrector.io

agent/internal/grpc/grpc.go
```diff
@@ -203,8 +202,17 @@ func grpcLoop(
 		if grpcConn.Client == nil {
 			client := agent.NewAgentClient(grpcConn.Conn)
 			grpcConn.SetClient(client)
+
+			publicKey, keyErr := config.GetPublicKey(string(appConfig.SecretPrivateKey))
+
+			if keyErr != nil {
+				log.Println("gprpc public key error")
```
polaroi8d, line 209: What is gprpc?

agent/internal/mapper/array_map.go
```diff
@@ -0,0 +1,19 @@
+package mapper
```
polaroi8d, line 1: Missing tests.

agent/pkg/crane/k8s/secret.go
```diff
@@ -0,0 +1,66 @@
...
+
+	result, err := cli.Apply(s.ctx, secrets, metav1.ApplyOptions{
+		FieldManager: s.appConfig.FieldManagerName,
+		Force:        s.appConfig.ForceOnConflicts,
+	})
+
+	if result != nil {
+		log.Println("Secret updated: " + result.Name)
+		if len(values) > 0 {
+			s.avail = append(s.avail, name)
```
polaroi8d, line 51: Append the values and at the end of the func, return with err? Is this correct, return with err, always?

protobuf/proto/agent.proto
```diff
@@ -83,6 +91,50 @@ message DeployRequest {
...
+    // runtime config type if given, magic can happen
+    optional RuntimeConfigType runtimeConfigType = 7;
+    message Expose {
+      bool public = 1; // if expose is needed
+      bool tls = 2;    // if tls is needed
+    }
+    /* exposure configuration*/
+    optional Expose expose = 8;
+
+    message ConfigContainer {
```
polaroi8d, line 124: Duplicated, already exists in the common proto file.

polaroi8d: With, `user`, `expose`...

web/crux-ui/src/components/products/versions/deployments/instances/edit-instance-json.tsx
```diff
@@ -0,0 +1,181 @@
...
+  return config
+}
+
+const reducer = (state: CompleteContainerConfig, action: EditImageJsonAction): CompleteContainerConfig => {
+  const { type } = action
+
+  if (type === 'config-change') {
+    return imageConfigToCompleteContainerConfig(state, action.config)
+  }
+  if (type === 'set-content') {
```
polaroi8d, line 72: switch or ifelse would be better, nope? 🤔

web/crux-ui/src/components/products/versions/deployments/instances/edit-instance-json.tsx
```diff
@@ -0,0 +1,181 @@
...
+      propsOnParseError(err)
+    },
+    [throttle, propsOnParseError],
+  )
+
+  useEffect(
+    () =>
+      dispatch({
+        type: 'config-change',
+        // config: props.config,
```
polaroi8d, line 161: Unsued code.

polaroi8d (changes requested):

.vscode/settings.json
nandor-magyar: Sure, *svn, we don't want them committed also don't want them to be indexed or considered in any way by vscode.

polaroi8d: So you will list all the file formats? :D

agent/internal/grpc/grpc.go
nandor-magyar: custom dyrectorio protocol :wink:

agent/pkg/crane/k8s/secret.go
nandor-magyar: Yes, if there was an error that surely propagates but in the end it is probably nil, err can be nil.

agent/internal/grpc/grpc.go
polaroi8d: love it

web/crux-ui/src/components/products/versions/deployments/instances/edit-instance-json.tsx
robot9706: Originall it was an if-else but the linter did not like it because each branch returns.

polaroi8d: Hmm, same as know, but okey.

.vscode/settings.json
nandor-magyar: Only those to be skipped by search and the IDE, this list is probably from some suggestion.

agent/pkg/builder/container/container_builder.go
```diff
@@ -48,9 +47,9 @@ type ContainerBuilder interface {
 	WithPostCreateHooks(hooks ...LifecycleFunc) ContainerBuilder
 	WithPreStartHooks(hooks ...LifecycleFunc) ContainerBuilder
 	WithPostStartHooks(hooks ...LifecycleFunc) ContainerBuilder
+	// Create() ContainerBuilder
```
robot9706, line 50: Why is this commented? 🐙

polaroi8d (approved):

agent/pkg/builder/container/container_builder.go
nandor-magyar: Accidents happened.

PR #100, opened 2022-08-22
add: dyo cli

c3ppc3pp:
Dyo command line utility, right now it generates a docker compose file with defined parameters.

Work in Progress

PR #155, opened 2022-09-12
fix(agent): dagent network creation condition corrected

nandor-magyar:
Using our builder in #100 turned out that we need better checks.


robot9706 (approved):
😎

balanceee (approved):
+1

PR #158, opened 2022-09-13
fix: prisma containerConfig, agent defaults

nandor-magyar:
Prisma did not succeed, because of value in the table, env variables added to examples.


polaroi8d (approved):
LGTM

c3ppc3pp (approved):
LGTM

PR #159, opened 2022-09-13
fix(web): deployment improvements

balanceee:
PR contains the following:

- Added an error toast to the create deployment screen which applies if the user wants to create more than one preparing deployment
- Fixed the issue when after you made a successful deployment the previous ones were not moved to obsolate status
- Fixed the issue when you were getting toast errors after making two simple product deployments each after

### DIFF
--- a/web/crux-ui/locales/en/versions.json
+++ b/web/crux-ui/locales/en/versions.json
@@ -27,5 +27,6 @@
   "addDeployment": "Add deployment",
   "addImage": "Add image",
 
-  "deploymentName": "Deployment name"
+  "deploymentName": "Deployment name",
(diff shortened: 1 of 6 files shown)


web/crux-ui/locales/en/versions.json
```diff
@@ -27,5 +27,6 @@
   "addDeployment": "Add deployment",
   "addImage": "Add image",
 
-  "deploymentName": "Deployment name"
+  "deploymentName": "Deployment name",
+  "preparingLimited": "You already have one preparing deployment!"
```
m8vago, line 31: alreadyHavePreparing

polaroi8d (approved):
LGTM, tested in local.

PR #160, opened 2022-09-14
fix(agent): extra key validation added 

polaroi8d:


agent/internal/config/secret.go
```diff
@@ -16,7 +17,17 @@ type ConfigFromFile string
 
 func (field *ConfigFromFile) SetValue(location string) error {
 	if location == "" {
-		return fmt.Errorf("field value can't be empty")
+		return fmt.Errorf("env private key file value can't be empty")
+	}
+
+	validKeyFile, err := regexp.MatchString("(.*?)\\.(key)$", location)
```
polaroi8d, line 23: I don't know, we can accept only .key files, @nandor-magyar?

robot9706 (approved):
lgtm 😎

agent/internal/config/secret.go
```diff
@@ -40,10 +40,15 @@ func checkGenerateKeys(location string) (string, error) {
...
-		}
+	if keyErr != nil {
+		return "", keyErr
+	}
+
+	if privateKeyObj == nil {
+		return "", fmt.Errorf("key file is nil: %v", location)
+	}
+
+	if privateKeyObj != nil && !privateKeyObj.IsExpired() {
```
nandor-magyar, line 51: This not-nil check is redundant this way, because of line 47.

nandor-magyar (approved):

PR #161, opened 2022-09-14
feat(web): removed deployment name

robot9706:
- Removed name from deployment.
- Renamed description to note.
- Added note popup to deployment tables.
- Added prefix and updatedAt columns to deployments table.


balanceee (commented):
Code looks good to me but i found some interesting things during testing:

- I can set the prefix as empty string and in this case the deployment fails
- If i try to create a deployment with a used prefix then the application is broke? (I got navigated to the status screen)
- On the deployment screen i cant distinguishes the deployments if they are in the same version and uses the same node. (Maybe we should show the prefix somewhere)
- Similiar to the webhook message. "Failed to deploy on node ${nodename}" is not enough i think

image
image

balanceee (approved):
LGTM

nandor-magyar (approved):

nandor-magyar (approved):

polaroi8d (approved):

PR #162, opened 2022-09-14
feat(crux-ui): add node status filter

balanceee:
**Added a node status filter to node list screen.**

PR also contains additonaly the following:

- Fix missing localization on registry screen.
- Add a toast error to the add deployment screen which applies when the user has no nodes

### DIFF
--- a/web/crux-ui/src/pages/nodes.tsx
+++ b/web/crux-ui/src/pages/nodes.tsx
@@ -6,10 +6,11 @@ import Filters from '@app/components/shared/filters'
 import PageHeading from '@app/components/shared/page-heading'
 import { ListPageMenu } from '@app/components/shared/page-menu'
 import { DyoHeading } from '@app/elements/dyo-heading'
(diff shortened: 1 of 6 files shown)


web/crux-ui/src/pages/nodes.tsx
```diff
@@ -96,7 +109,36 @@ const NodesPage = (props: NodesPageProps) => {
       {!creating ? null : <EditNodeCard className="mb-2" submitRef={submitRef} onNodeEdited={onCreated} />}
       {filters.items.length ? (
         <>
-          <Filters setTextFilter={it => filters.setFilter({ text: it })} />
+          <DyoCard className="flex flex-col p-8">
```
bvektor, line 112: Wouldn't the Filters component with DyoChips inside be better? Like on the products page

polaroi8d (approved):
LGTM, tested in local.

nandor-magyar (approved):

PR #168, opened 2022-09-16
fix(crux-ui): version hint

peterambrus3:
fix(crux-ui): Version semantic hint
fix(crux-ui): Large gap on the right side of the main layout
fix(crux-ui): Not consistent breadcrumb header positions

### DIFF
--- a/web/crux-ui/locales/en/versions.json
+++ b/web/crux-ui/locales/en/versions.json
@@ -1,7 +1,7 @@
 {
   "versionsName": "Products - {{product}} - Versions - {{name}}",
 
-  "tips": "Create a new version for your product that you can use for different purposes. Rolling versions are ideal for continuous delivery, while Incremental ones are useful for production.",
+  "tips": "Create a new version for your product. You can select a default version, from which other versions will inherit images and configuration. dyrector.io is the most suitable for semantic versioning, since it's the de facto way of version control. We recommend using that.",
   "increaseTips": "By increasing the version of your Product, you're able to create new releases of your already existing Product with new images or substitute included ones with a different tag.",
   "new": "New version",
   "default": "Default",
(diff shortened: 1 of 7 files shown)


web/crux-ui/src/components/products/versions/edit-version-card.tsx
```diff
@@ -81,6 +81,21 @@ const EditVersionCard = (props: EditVersionCardProps) => {
     submitRef.current = formik.submitForm
   }
 
+  const checkVersionHint = (value: string) => {
+    const err = getValidationError(nameRule, value)
+    if (!versionRegex.test(value) && !err) {
```
robot9706, line 86: Wouldn't be easier to use yup to validate the regex?

robot9706 (changes requested):

web/crux-ui/locales/en/versions.json
```diff
@@ -1,7 +1,7 @@
 {
   "versionsName": "Products - {{product}} - Versions - {{name}}",
 
-  "tips": "Create a new version for your product that you can use for different purposes. Rolling versions are ideal for continuous delivery, while Incremental ones are useful for production.",
+  "tips": "Create a new version for your product. You can select to default a version other versions will inherit images and configurations from. dyrector.io is the most suitable for semantic versioning, since it's the de facto way of version control. We recommend using that.",
```
robot9706, line 4: `You can select a default version, from which other versions will inherit images and configuration` ?

robot9706 (changes requested):

web/crux-ui/locales/en/versions.json
peterambrus3: i just followed the instructions, but i like your version

polaroi8d: @robot9706 suggestion looks better for me.

robot9706 (approved):
LGTM 😎

nandor-magyar (approved):

PR #169, opened 2022-09-16
feat: add init-containers to crane

nandor-magyar:
Arbitrary init containers can be defined with this for imports, seeding etc.


web/crux-ui/src/server/crux/mappers/image-mappers.ts
```diff
@@ -21,6 +21,7 @@ import {
   restartPolicyToJSON,
   Volume,
 } from '@app/models/grpc/protobuf/proto/common'
+import { InitContainer } from './../../../models/grpc/protobuf/proto/common'
```
robot9706: Use @app import.

agent/pkg/crane/k8s/deployment.go
```diff
@@ -81,7 +81,7 @@ func (d *deployment) deployDeployment(p *deploymentParams) error {
 					d.appConfig.KeyIssuer:  p.issuer,
 				}).WithSpec(
 					corev1.PodSpec().WithContainers(containerConfig).
-						WithInitContainers(getInitContainers(p.containerConfig, d.appConfig)...).
+						WithInitContainers(getInitContainers(p.containerConfig, d.appConfig, p)...).
```
robot9706, line 84: If `p` is passed to this function then passing `p.containerConfig` could be simplyfied just by passing `p` and then getting `containerConfig` later in the function.

robot9706 (changes requested):

robot9706 (approved):
LGTM 😎

c3ppc3pp (approved):
:+1:

PR #170, opened 2022-09-16
feat: add secret volume to k8s manifests

nandor-magyar:
Recent changes to secret management and end-to-end encrypted secrets need us to allocate a volume in the cluster upon the installation of crane.


robot9706 (approved):
😎

balanceee (approved):
+1

PR #178, opened 2022-09-16
fix: deploying on unreachable node

m8vago:
When clicking on the deploy button while the target node is unreachable, the app shows a toast now instead of navigation to the deploy page.


web/crux-ui/src/pages/products/[productId]/versions/[versionId]/deployments/[deploymentId].tsx
```diff
@@ -134,6 +134,11 @@ const DeploymentDetailsPage = (props: DeploymentDetailsPageProps) => {
       return
     }
 
+    if (deployment.node.status !== 'running') {
```
Comment (polaroi8d) [inline web/crux-ui/src/pages/products/[productId]/versions/[versionId]/deployments/[deploymentId].tsx:137]: I think we can check this at the beginning of the `onDeploy()` to reach the fail fast approach, what do you think?

Comment (m8vago) [reply web/crux-ui/src/pages/products/[productId]/versions/[versionId]/deployments/[deploymentId].tsx:137]: sure!

robot9706 (approved):
LGTM 😎

polaroi8d (approved):
LGTM

PR #179, opened 2022-09-19
feat(web): registry delete check & toast

robot9706:
When deleting a registry check if any image uses it, if so show a toast instead of deleting.


polaroi8d (approved):
LGTM

balanceee (approved):
LGTM

PR #186, opened 2022-09-20
fix(crux-ui): use regex on deployment prefix validation

balanceee:


web/crux-ui/src/validation.ts
```diff
@@ -138,17 +138,21 @@ export const createVersionSchema = updateVersionSchema.concat(
...
-  note: yup.string(),
-})
-
 export const updateDeploymentSchema = yup.object().shape({
   note: yup.string(),
-  prefix: yup.string().trim().required(),
+  prefix: yup
+    .string()
+    .trim()
+    .matches(/^[a-z0-9]([-a-z0-9]*[a-z0-9])?$/) // RFC 1123
```
polaroi8d, line 146: Why do we need to match the prefix for this?

polaroi8d (approved):

PR #195, opened 2022-09-21
feat(web): added root path to agent script

robot9706:
Added an input to the agent script generator card to define the root path for Docker agents.


web/crux-ui/src/components/nodes/dyo-node-setup.tsx
```diff
@@ -33,6 +33,7 @@ const DyoNodeSetup = (props: DyoNodeSetupProps) => {
     node.install ? expiresIn(new Date(node.install.expireAt)) : null,
     () => onNodeInstallChanged(null),
   )
+  const [rootPath, setRootPath] = useState<string>('')
```
nandor-magyar, line 36: It could take the default path `/srv....` - default value.

robot9706: It's different on linux and mac 😅 I could also just use null.

nandor-magyar: That is why it is important to be explicit about it, even a select of reasonable defaults could guide the user a bit.

web/crux/install-docker.sh.hbr
```diff
@@ -10,10 +10,10 @@ fi
 # Setup the default root folder for volumes
 if [ $(uname -s) = 'Darwin' ]; then
   echo "Detected: Mac OS X"
-  ROOT_FOLDER=$HOME'/Library/Dyrectorio/srv/dagent'
+  ROOT_FOLDER=$HOME'{{#if rootPath}}{{{rootPath}}}{{else}}/Library/Dyrectorio/srv/dagent{{/if}}'
```
polaroi8d, line 13: Remove `$HOME` too.

polaroi8d (approved):

nandor-magyar (approved):

PR #197, opened 2022-09-22
feat(cli): check docker (and podman) requirements

c3ppc3pp:
So the basic Idea is, we check a bunch of reqs for docker, so when we open a client or use the containerbuilder we wont get unhelpful errors, CLI will tell where it failed. I added some podman checks too, since our stuff requires netavark networkbackend in podman (specifically communication between containers and proper DNS resolutions are only possible with this).


cli/config_file.go
```diff
@@ -155,6 +162,91 @@ func SettingsFileReadWrite(state *Settings) *Settings {
 	return settings
 }
 
+// Check prerequisites
+func CheckRequirements() {
+	// getenv
+	envVarValue := os.Getenv("DOCKER_HOST")
+
+	if envVarValue == "" {
+		log.Fatalf("DOCKER_HOST environmental variable is empty or not set")
```
nandor-magyar, line 171: I have it empty by default, are you sure it is reason to panic?

cli/config_file.go
```diff
@@ -155,6 +162,91 @@ func SettingsFileReadWrite(state *Settings) *Settings {
...
+		log.Fatalf("unknown init binary")
+	}
+}
+
+func PodmanInfo() {
+	cmd := exec.Command("podman", "info", "--format", "{{.Host.NetworkBackend}}")
+
+	stderr, err := cmd.StderrPipe()
+	if err != nil {
+		log.Fatalf("error: %v", err)
```
nandor-magyar, line 214: `podman check error` would be better,  error is not really descriptive, here and docker checks too.

cli/config_file.go
c3ppc3pp: Yes, since this is how docker's SDK works afaik. You cannot really assume the DOCKER_HOST will be a default unix:///var/run/dokcer.sock, since docker can use contexts (remote tcp/ssh sockets, etc), and also socket can be in other places too.

In my opinion, I'd like to force out a properly set DOCKER_HOST env var, than encounter issues later on. Having the default docker socket which isn't working properly for some reason, like my case, only root can access to that socket, could possibly misguide some users because of the misleading error messages.

Should we assume the DOCKER_HOST should be unix:///var/run/dokcer.sock if empty?

cli/config_file.go
c3ppc3pp: Indeed, but it is also a good idea to append useful bits of information everywhere in error messages, since lot of places using this "format". Thanks!

cli/config_file.go
m8vago: i think it would be a nice tradeoff, if we could write a highlighted warning message like: `DOCKER_HOST environmental variable is empty or not set. Using unix:///var/run/dokcer.sock as default`

polaroi8d: And set default to the user.

polaroi8d: After playing with DOCKER_HOST, IMHO just drop a warning message as @m8vago said would be okey for me.

cli/config_file.go
c3ppc3pp: fixed, thanks!

cli/config_file.go
c3ppc3pp: I rearranged it a bit, now it checks the existence of DOCKER_HOST, if it doesn't it will jump straight to connection phase while printing a yellow ERROR message, but when it's present it will try to parse it, and check validity before trying to connect to the socket.

robot9706 (approved):
LGTM 😎

cli/config_file.go
```diff
@@ -155,6 +162,93 @@ func SettingsFileReadWrite(state *Settings) *Settings {
...
+		if socketurl.Host != "" {
+			log.Fatalf("error: DOCKER_HOST variable shouldn't have host")
+		}
+
+		if socketurl.Scheme != "unix" {
+			log.Fatalf("error: DOCKER_HOST variable should contain a valid unix socket")
+		}
+	} else {
+		// We cannot assume unix:///var/run/docker.sock on Mac/Win platforms, we let Docker SDK does its magic :)
+		log.Println("\033[33mwarning: DOCKER_HOST environmental variable is empty or not set.")
```
polaroi8d, line 185: image

Stop coloring every line ending or just do it in one Println.

polaroi8d (approved):
Added a minor issue. Sidenote: Our branching policy does not consist to add the scope to the branch name, referring to `/cli`.

cli/config_file.go
c3ppc3pp: now both lines have start/stop characters in them regarding to coloring

cli/config_file.go
c3ppc3pp: fixed

cli/config_file.go
```diff
@@ -155,6 +162,93 @@ func SettingsFileReadWrite(state *Settings) *Settings {
...
+		log.Fatalf("unknown init binary")
+	}
+}
+
+func PodmanInfo() {
+	cmd := exec.Command("podman", "info", "--format", "{{.Host.NetworkBackend}}")
+
+	stderr, err := cmd.StderrPipe()
+	if err != nil {
+		log.Fatalf("error - podman check stderr pipe error: %v", err)
```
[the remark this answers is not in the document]
c3ppc3pp: error messages were sorted out regarding the redundancy

PR #199, opened 2022-09-26
feat(crux-ui): add chip filters

bvektor:
- add chip filters to deployments, registries and nodes list pages


web/crux-ui/src/elements/dyo-chips.tsx
```diff
@@ -1,27 +1,31 @@
 import assert from 'assert'
 import clsx from 'clsx'
+import useTranslation from 'next-translate/useTranslation'
 import { useState } from 'react'
 
 interface DyoChipsProps<T> {
   className?: string
   key?: React.Key
   choices: readonly T[]
+  isFilter?: boolean
```
m8vago, line 10: Let's be explicit about the implication of this prop. Rename it to addAllOption or something

web/crux-ui/src/hooks/use-filters.ts
```diff
@@ -98,7 +98,7 @@ export type EnumFilter<TEnum> = {
 export const enumFilterFor =
   <Item, Enum>(propertiesOf: PropertyValuesOf<Item>): FilterFunction<Item, EnumFilter<Enum>> =>
   (items: Item[], filter: EnumFilter<Enum>) => {
-    if (filter.enum === undefined) {
+    if (filter?.enum === undefined || filter?.enum === null) {
```
m8vago, line 101: is there a case where the filter is not declared?
if no then:
```
if (!filter.enum)
```

web/crux-ui/src/hooks/use-filters.ts
```diff
@@ -51,14 +51,14 @@ export const useFilters = <Item, Filter>(options: UseFiltersOptions<Item, Filter
...
-  text: string
+  text?: string
 }
 
 export const textFilterFor =
   <Item>(propertiesOf: PropertyValuesOf<Item>): FilterFunction<Item, TextFilter> =>
   (items: Item[], filter: TextFilter) => {
     const text = filter?.text?.trim().toLowerCase()
-    if (text && text.length < 1) {
+    if (!text || (text && text.length < 1)) {
```
m8vago, line 61: dont need the `text &&` part

web/crux-ui/src/pages/products.tsx
```diff
@@ -18,15 +18,15 @@ import useTranslation from 'next-translate/useTranslation'
 import { useRouter } from 'next/dist/client/router'
 import { useRef, useState } from 'react'
 
-export const PRODUCT_TYPE_FILTER_VALUES = ['simple', 'complex', 'all'] as const
+export const PRODUCT_TYPE_FILTER_VALUES = [null, 'simple', 'complex'] as const
```
polaroi8d, line 21: Can we use `unknow` instead of `null`?

robot9706:
On the version deployments screen the status filter is different from the deployments list:
image
Is this on purpose?

web/crux-ui/src/elements/dyo-filter-chips.tsx
```diff
@@ -0,0 +1,16 @@
+import useTranslation from 'next-translate/useTranslation'
+import DyoChips, { DyoChipsProps } from './dyo-chips'
+
+interface DyoFilterChipsProps<T> extends DyoChipsProps<T> {
+  addAllOption?: boolean
```
m8vago, line 5: is this ever false?

web/crux-ui/src/hooks/use-filters.ts
m8vago: I would set all as a default value then, so we don't have to worry about the filter being null

m8vago: nevermind i got it

m8vago: however:

```
const { enum } = filter
if (!enum)
```

robot9706 (approved):
LGTM 😎

polaroi8d (approved):

PR #203, opened 2022-09-27
fix: check directory for secret keys

polaroi8d:
Fix contains the followings:
- We have to check the `EISDIR` error and added the missing error check from `os.ReadFile()`
- Rename `location` to `path`

### DIFF
--- a/agent/internal/config/secret.go
+++ b/agent/internal/config/secret.go
@@ -8,18 +8,19 @@ import (
 	"fmt"
 	"log"
 	"os"
+	"syscall"
 
 	"github.com/ProtonMail/gopenpgp/v2/crypto"
 )
 
 type ConfigFromFile string
 
-func (field *ConfigFromFile) SetValue(location string) error {
-	if location == "" {
+func (field *ConfigFromFile) SetValue(path string) error {
+	if path == "" {
 		return fmt.Errorf("env private key file value can't be empty")
 	}
 
-	key, err := checkGenerateKeys(location)
+	key, err := checkGenerateKeys(path)
 	if err != nil {
 		return err
 	}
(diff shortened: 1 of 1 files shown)


agent/internal/config/secret.go
```diff
@@ -28,37 +29,43 @@ func (field *ConfigFromFile) SetValue(location string) error {
 	return nil
 }
 
-func checkGenerateKeys(location string) (string, error) {
-	log.Printf("Checking key file: %v\n", location)
-	file, err := os.ReadFile(location) //#nosec G304 -- location comes from an env
+func checkGenerateKeys(path string) (string, error) {
+	log.Printf("Checking key file: %v\n", path)
+	readFile, err := os.ReadFile(path) //#nosec G304 -- path comes from an env
```
nandor-magyar, line 34: minor: readFile sounds like a function, fileContent or similar would be a bit better.

agent/internal/config/secret.go
```diff
@@ -8,18 +8,19 @@ import (
...
+	"syscall"
 
 	"github.com/ProtonMail/gopenpgp/v2/crypto"
 )
 
 type ConfigFromFile string
 
-func (field *ConfigFromFile) SetValue(location string) error {
-	if location == "" {
+func (field *ConfigFromFile) SetValue(path string) error {
```
nandor-magyar, line 18: Using `path` as variable name would conflict/shadow the `path` package from stdlib, it is better be avoided, it is painful when it happens.

agent/internal/config/secret.go
```diff
@@ -28,37 +29,43 @@ func (field *ConfigFromFile) SetValue(location string) error {
...
 }
 
-func checkGenerateKeys(location string) (string, error) {
-	log.Printf("Checking key file: %v\n", location)
-	file, err := os.ReadFile(location) //#nosec G304 -- location comes from an env
+func checkGenerateKeys(path string) (string, error) {
+	log.Printf("Checking key file: %v\n", path)
+	readFile, err := os.ReadFile(path) //#nosec G304 -- path comes from an env
+
+	if errors.Is(err, syscall.EISDIR) {
```
robot9706, line 36: Chaining the if-s (if-else if-else) would look better than the current if if-else. Opinions?

robot9706 (changes requested):

nandor-magyar (approved):

agent/internal/config/secret.go
polaroi8d: Talked in person.

robot9706 (approved):
LGTM 😎

agent/internal/config/secret.go
polaroi8d: Fixed.

agent/internal/config/secret.go
polaroi8d: Fixed.

PR #211, opened 2022-09-28
feat(web): use default container list on deployment screen

balanceee:
From now, on the deployment's log screen you can always see your containers (even if they are deleted). Previously when you started a deplyoment the container's list were uploaded continuously. Now only their status are updated. Also the prefix has been removed from the container's name in case of docker.

### DIFF
--- a/web/crux/src/app/node/node.service.ts
+++ b/web/crux/src/app/node/node.service.ts
@@ -1,6 +1,5 @@
 import { Injectable, Logger } from '@nestjs/common'
 import { Observable } from 'rxjs'
-import { PreconditionFailedException } from 'src/exception/errors'
 import {
   AccessRequest,
   ContainerStateListMessage,
@@ -21,6 +20,7 @@ import {
 import PrismaService from 'src/services/prisma.service'
 import DomainNotificationService from 'src/services/domain.notification.service'
 import { BaseMessage } from 'src/domain/notification-templates'
+import { PreconditionFailedException } from 'src/exception/errors'
 import AgentService from '../agent/agent.service'
 import TeamRepository from '../team/team.repository'
 import NodeMapper from './node.mapper'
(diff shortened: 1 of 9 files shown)


web/crux/src/app/node/node.service.ts
```diff
@@ -175,19 +176,79 @@ export default class NodeService {
...
+        },
+      },
+    })
+
+    const defaultContainerState = {
+      prefix: request.prefix,
+      data: defaultImages.instances.map(
+        it =>
+          ({
+            name: `/${request.prefix}-${it.image.name}`,
```
robot9706, line 206: Afaik the container name is `image.config.name`

web/crux/src/app/node/node.service.ts
```diff
@@ -175,19 +176,79 @@ export default class NodeService {
     return await this.agentService.getNodeEventsByTeam(request.id)
   }
 
-  handleWatchContainerStatus(request: WatchContainerStateRequest): Observable<ContainerStateListMessage> {
+  async handleWatchContainerStatus(
+    request: WatchContainerStateRequest,
+  ): Promise<Observable<ContainerStateListMessage>> {
     this.logger.debug(`Opening container status channel for prefix: ${request.nodeId} - ${request.prefix}`)
```
peterambrus3, line 182: Can you extend the log with the newly added deploymentID?

balanceee: Yes. Sure

web/crux/src/app/node/node.service.ts
balanceee: They are the same. Maybe its redundant?

robot9706: Not sure then 🤷‍♀️

web/crux-ui/tailwind.config.js
```diff
@@ -13,6 +13,7 @@ module.exports = {
                 'dyo-purple': '#540ba7',
                 'dyo-purple-light': '#e1ccf8',
                 'dyo-blue': '#00cfe8',
+                'dyo-gray': '#d0d2d6',
```
m8vago, line 16: dyo-gray to bright

m8vago (approved):

peterambrus3 (approved):
🤠

Issue #219, opened 2022-09-30
Upgrade Prisma Client to 4.4.0


polaroi8d:
As the title says upgrade the Prisma package in `web/crux` service.

Panda-Abhisek:
I am a beginner at contributing.
Can you guide me in this?

polaroi8d:
Sorry @Panda-Abhisek but it's already done by Kit. Reference: Reference #226

Panda-Abhisek:
OK

PR #225, opened 2022-09-30
refactor(crux): use @updatedAt in prisma schema

### LINKED ISSUE #208: Use Prisma @updateAt signature instead of manually doing
We have a few `updateAt` columns in our database schema. To avoid boilerplate codes Prisma has a sophisticated solution for this particular problem. They made an `@updateAt` signature: _automatically stores the time when a record was last updated. If you do not supply a time yourself, the Prisma Client will automatically set the value for fields with this attribute._

Prisma Reference documentation: https://www.prisma.io/docs/reference/api-reference/prisma-schema-reference#updatedat

Kit-p:
Replaces `@default(now())` with `@updatedAt` and removes all `updatedAt` in code when updating records.

> Important: This PR introduces one unwanted behavior, where new records created may have the `updatedAt` field not matching `createdAt` if wherever the database is hosted uses a different system time than the client, due to https://github.com/prisma/prisma/issues/12572.

The above unwanted behavior is tracked in #219.

Feedback is welcome!

Closes #208.

### DIFF
--- a/web/crux/prisma/migrations/20220929162700_use_updated_at_attribute/migration.sql
+++ b/web/crux/prisma/migrations/20220929162700_use_updated_at_attribute/migration.sql
@@ -1,0 +1,26 @@
+-- AlterTable
+ALTER TABLE "Deployment" ALTER COLUMN "updatedAt" DROP DEFAULT;
+
+-- AlterTable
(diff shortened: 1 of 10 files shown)


robot9706 (approved):
LGTM 🚀

polaroi8d (approved):
LGTM, good job!

Issue #209, opened 2022-09-28
Check GRPC_TOKEN environment variable is a JWT Token


polaroi8d:
In `agent/internal/config/config.go` we store our configuration defaults. We use `GRPC_TOKEN` which is a JWT Token, but right now it's a simple string. Please create a validation for it to check if it's a JWT token.

nandor-magyar:
Custom parsing and checks are done in `config/secret.go`.

trkohler:
hello! can I work on this issue for hacktoberfest participation?

nandor-magyar:
@trkohler sure, go ahead, feel free to work on it, don't hesitate to ask if you have any questions, you can find us on [Discord](https://discord.gg/n5ASdrz3pZ).

> Custom parsing and checks are done in `config/secret.go`.

I meant that the given file is a good starting point regarding the issue above. I would create a type and would put related code to a new file like `jwt.go` - similarly that is done with the type `ConfigFromFile`.

trkohler:
Hello! Hope I've did as you've counted I would. PR is ready for review.

PR #228, opened 2022-10-01
feat: JWT token validation

trkohler:
resolve https://github.com/dyrector-io/dyrectorio/issues/209


polaroi8d:
[GPG commit signature](https://github.com/dyrector-io/dyrectorio/blob/develop/CONTRIBUTING.md#commit-signatures) is missing and the wrong commit message. Please read CONTRIBUTING.md for more information.

trkohler:
I signed commit with GPG key with my email but for some reason it tells `No user is associated with the committer email.` even despite the fact that it's the email I use for my GH account 😔

agent/internal/config/jwt.go
```diff
@@ -0,0 +1,77 @@
...
+	if err != nil {
+		return err
+	}
+
+	*field = token
+	return nil
+}
+
+func ValidateAndCreateJWT(unvalidatedToken string) (JWTToken, error) {
+	splitted := strings.Split(unvalidatedToken, ".")
```
nandor-magyar, line 45: `splitted` is not a correct English word, use a correct one instead.

agent/pkg/crane/crane.go
```diff
@@ -31,10 +31,6 @@ func Serve(cfg *config.Configuration) {
 	grpcToken := cfg.GrpcToken
 	grpcInsecure := cfg.GrpcInsecure
 
-	if grpcToken == "" {
```
nandor-magyar, line 34: This is safe to remove, only if in `config.go` it is marked as required.

agent/pkg/dagent/dagent.go
```diff
@@ -18,10 +18,6 @@ func Serve(cfg *config.Configuration) {
 	grpcToken := cfg.GrpcToken
 	grpcInsecure := cfg.GrpcInsecure
 
-	if grpcToken == "" {
```
nandor-magyar, line 21: Same as above.

agent/internal/config/jwt_test.go
```diff
@@ -0,0 +1,52 @@
...
+	valid               bool
+}
+
+func (j jwtTest) run(t *testing.T) {
+	_, err := ValidateAndCreateJWT(j.jwtTokenStringified)
+	if err != nil && j.valid {
+		t.Errorf("%v", err)
+	}
+	if err == nil && !j.valid {
+		t.Errorf("expected error")
```
nandor-magyar, line 17: If the error is expected, it could simply be t.Log,no?

agent/internal/config/config.go
```diff
@@ -20,10 +20,10 @@ type CommonConfiguration struct {
 	GrpcKeepalive        time.Duration `yaml:"grpcKeepalive"         env:"GRPC_KEEPALIVE"          env-default:"60s"`
 	Debug                bool          `yaml:"debug"                 env:"DEBUG"                   env-default:"false"`
 	// TODO(c3ppc3pp): custom setter to validate input
```
nandor-magyar, line 22: The comment belongs here, custom setter being done, this is safe to be removed.

nandor-magyar (changes requested):
Amazing, good job! :sunglasses: 
The implementation looks good, minor things left to polish. :+1:

agent/pkg/crane/crane.go
trkohler: I make validation for grpcToken to not be an empty string 

```
if unvalidatedToken == "" {
		return fmt.Errorf("grpc jwt token can not be unset")
	}
```

after that the struct is being created, so grpcToken won't be "" in any case, isn't it?

agent/internal/config/jwt_test.go
trkohler: changed accordingly.

agent/internal/config/config.go
trkohler: cleaned out.

agent/internal/config/jwt.go
trkohler: [changed to `split` ](https://preply.com/en/question/past-tense-of-split)

agent/pkg/crane/crane.go
nandor-magyar: It is correct. That check too could be removed, if the required option is set.
Consult the docs here https://github.com/ilyakaznacheev/cleanenv `env-required`
The error message is mixed anyway, the type is for JWT token, it has nothing to do with gRPC directly, for this reason I prefer that solution, however I cannot recall how good are those error messages provided by the cleanenv package. Let's see how is it that way.

trkohler: in their doc they show almost the same [example when field is of type custom struct](https://github.com/ilyakaznacheev/cleanenv#custom-value-setter)

also [from their examples with the code ](https://github.com/ilyakaznacheev/cleanenv/blob/35edfee562627f73280939ac560966bcdd439cda/example/custom_setter/main.go#L39)

env-required is not compatible with custom structs. It doesn't know how to transform env value to custom struct (but the setter knows)

i've changed error messages. maybe it can work out that way?

agent/internal/config/config.go
trkohler: not sure on that one: i've removed the comment, anything else which I may miss?

agent/internal/config/jwt.go
```diff
@@ -0,0 +1,77 @@
...
+	Signature     string
+	InitialString string
+}
+
+type JWTHeader struct {
+	Alg string `json:"alg"`
+	Typ string `json:"typ,omitempty"`
+}
+
+type JWTPayload struct {
```
c3ppc3pp, line 22: If I recall correctly we do not use fields like this, as we are using JWTs for an initial auth (between agent and backend). I'll get back to you tomorrow.

trkohler: yes, the code doesn't utilize those, only `initialString` with the separate jwt library when parse with claims.

those types are made specifically for validation to ensure the JWT token has proper anatomy.

trkohler: [the codebase currently works with the string](https://github.com/dyrector-io/dyrectorio/pull/228/files#diff-e0c54bda26f4cc7eaa531f24a5237487e30bec13a10d76f23ef5662eb6c1b794R77) and [there is an external library JWT parsing](https://github.com/dyrector-io/dyrectorio/pull/228/files#diff-e0c54bda26f4cc7eaa531f24a5237487e30bec13a10d76f23ef5662eb6c1b794R64) so my types are not doing much except some explicitness and validation purposes.

nandor-magyar (changes requested):
Sorry for the hassle, instead of what is done here, I propose another solution: validate with the official JWT package.
https://pkg.go.dev/github.com/golang-jwt/jwt/v4

It is good for the long-run, but for that we would need to have a deeper look and that is out of scope for this PR.
This would be a good start and would satisfy the nice tests you've added.

```go
  jwtParser := jwt.Parser{}
  token, _, err := jwtParser.ParseUnverified(unvalidatedToken, &jwt.StandardClaims{})
```

Could be used in the function ValidateAndCreateJWT to keep the same functionality.

Edit: fixed url.

trkohler:
>  I propose another solution: validate with the official JWT package.

I've changed the validation to the way you suggested @nandor-magyar and all un-necessary types and logic are gone.

agent/internal/config/jwt.go
```diff
@@ -0,0 +1,36 @@
+package config
+
+import (
+	"fmt"
+	"github.com/form3tech-oss/jwt-go"
```
nandor-magyar, line 5: It was not this package that I've suggested, if you check the repository this is archived and read-only. Use the one mentioned in the comment.

agent/internal/config/jwt.go
```diff
@@ -0,0 +1,36 @@
...
+}
+
+func ValidateAndCreateJWT(unvalidatedToken string) (ValidJWT, error) {
+
+	token := ValidJWT{
+		Token: unvalidatedToken,
+	}
+
+	jwtParser := jwt.Parser{}
+	_, _, err := jwtParser.ParseUnverified(unvalidatedToken, &jwt.StandardClaims{})
```
nandor-magyar, line 33: Here is an example token for local development. Add validation if the payload contains the all fields as in the token, every field is required (iat, iss, sub).

`eyJhbGciOiJIUzI1NiIsInR5cCI6IkpXVCJ9.eyJpYXQiOjE2NjQ4NzQxMDM5NzQsImlzcyI6IjE3Mi4xNy4wLjE6NTAwMCIsInN1YiI6ImIzZmYwMmU5LWU4MDUtNDhjYy1hYzNkLWRhMDVhZTEwMjNkZiJ9.qVlU38JwsLFrRf85PMhMFjZ7Jm4wyxWovDCa7nPNV0g`

nandor-magyar (commented):
See my comments.

agent/internal/config/jwt.go
trkohler: I've used the same package which is already being used in your codebase there `agent/internal/grpc/grpc.go`

The link you provided does not work, it gives 400 status code

I'm assuming you meant that one package? https://pkg.go.dev/github.com/golang-jwt/jwt
I will use it, it seems to be alive and have a lot of users

agent/internal/config/jwt.go
trkohler: `jwtParser.ParseUnverified` checks all the anatomy of the JSON token and in case of some missing values which should be required -- it returns a value which states that. Why to introduce additional checks?

agent/internal/config/jwt.go
trkohler: do you want legacy jwt library to be remained in the codebase and me - introducing a new library only in the value setter & validator? or may I refactor the part with the old jwt library as well?

trkohler: I've made the validation with the new library, but have not touched the parts in the existing codebase.

Also I've slightly changed the tests because the behaviour of the library is different from what I've had before.

nandor-magyar: It would be nice if you could replace it everywhere.

agent/internal/config/jwt.go
nandor-magyar: The first check is to tests that the construct, JWT is valid, if it is valid we need crucial parts to be defined in the payload, those have impact on agent behavior, if they are missing it is an invalid configuration, as stated, every field of those listed above are strictly required, therefore we should check for existence on startup.

polaroi8d: @trkohler payload validation missing. How we can help with this task to you?

polaroi8d:
@trkohler any update from your side?

trkohler:
i won't process with that, closing the PR. Thanks for your patience.

Issue #221, opened 2022-09-30
Add new builder: container exec


nandor-magyar:
Based on our container [builder](https://github.com/dyrector-io/dyrectorio/tree/develop/agent/pkg/builder/container)

TODO: Add a new builder named ExecBuilder, into the same package into a file `exec_builder.go`.

The builder should be created new a NewExecBuilder(containerID string) function.
Use the builder to set any given parameter and .Start() should return the output / result of the executed command.

Use the API as a reference.
https://docs.docker.com/engine/api/v1.41/#tag/Exec/operation/ContainerExec

Golang package:
https://pkg.go.dev/github.com/docker/docker/client#Client.ContainerExecCreate

PankhudiB:
Can I work on this issue under Hacktoberfest '22 ?

PR #230, opened 2022-10-03
feat: introduce container exec builder to run exec command on container

PankhudiB:
Add new builder: container exec #221

### DIFF
--- a/agent/pkg/builder/container/exec_builder_test.go
+++ b/agent/pkg/builder/container/exec_builder_test.go
@@ -1,0 +1,45 @@
+//go:build integration
+// +build integration
+
+package container_test
+
+import (
+	"context"
+	"testing"
+
+	"github.com/stretchr/testify/assert"
+
+	containerbuilder "github.com/dyrector-io/dyrectorio/agent/pkg/builder/container"
+)
+
+func TestExecBuilder(t *testing.T) {
+	logger := &testLogger{
+		test: t,
+	}
+
+	containerBuilder := containerbuilder.NewDockerBuilder(context.Background()).
+		WithName("test").
+		WithImage("alpine:latest").
+		WithCmd([]string{"/bin/sh", "-c", "sleep 5"})
+
+	dockerContainerBuilder := containerBuilder.Create()
+	containerID := dockerContainerBuilder.GetContainerID()
+
+	_, err2 := dockerContainerBuilder.Start()
+	if err2 != nil {
+		panic(err2.Error())
+	}
+	defer builderCleanup(containerBuilder)
+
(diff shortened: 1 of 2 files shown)


PankhudiB:
I am not sure how to go about writing test for exec. We can start a container but then unless we add a sleep or some way to keep the container running, we cant run Exec on it.  If you can suggest something else I can incorporate.

nandor-magyar (commented):
Good & quick job!

Our dockerContainerBuilder has a small design flaw, that would better not to be inherited.

The issue:
```go
var _ containerbuilder.ContainerBuilder = (*containerbuilder.DockerContainerBuilder)(nil)
```
It will be highlighted as invalid because it is.

To avoid this problem, the builder should build a struct that has behavioral features.

In case of ContainerExec:
- a new struct named Exec, that contains the result of Create
  - create becomes the final function of the builder
  - create can now return `(Exec,error)`
- move the function `Start` to the new type: Exec

Regarding tests:
The interface check above could be part of a test.
To verify exec behavior it is better to start a service type container eg. nginx and terminate it after all tests are completed.
You can do such things using TestMain.

Let me know if you have any questions.

PankhudiB:
If we maintain `Start()` as method for `Exec`, 
it will need us to maintain the `docker client` as state of Exec in order to eventually do 
`dockerClient.ContainerExecStart` in Start()

[ Which we can set during `Create() (Exec,error)` ]
I can do that but felt like an encapsulation mis-fit to me, hence wanted to confirm.

agent/pkg/builder/container/exec_builder_test.go
```diff
@@ -29,17 +31,23 @@ func TestExecBuilder(t *testing.T) {
...
+		WithCmd([]string{"export"})
 
-	success, err := execBuilder.Create().Start()
+	exec, execCreationErr := execBuilder.Create()
 
-	assert.Nil(t, err)
-	assert.True(t, success)
-	assert.True(t, logger.gotMessage)
+	if execCreationErr != nil {
+		panic(execCreationErr.Error())
```
nandor-magyar, line 46: Use t.Fatal instead of panic in tests.

nandor-magyar (commented):
It looks good. Could you please elaborate why do think there is an encapsulation misfit?

PankhudiB:
> It looks good. Could you please elaborate why do think there is an encapsulation misfit?

After refactoring in local , I realised what you are suggesting would be correct modelling 😅
the Exec now can have methods other than Start.

nandor-magyar (approved):
Good job! LGTM!

c3ppc3pp (approved):
Looks really neat! GJ!

polaroi8d (approved):

Issue #241, opened 2022-10-05
Write unit tests for agent and we'll send you an awesome dyrectorio swag cap


polaroi8d:
Right now `/agent` service code coverage is around 8.67% - at events such as Hacktoberfest, it's a good first issue to write unit tests around the code base. You can add tests to `/cli` if you're a black belt unit test writer.

We know unit test writing isn't the most exciting task, so we'd like to motivate our contributors. So if you add more than 10% coverage in the golang side of the repository, we'll send you an awesome dyrectorio swag cap. 

You can find codecov agent/ coverage: https://app.codecov.io/gh/dyrector-io/dyrectorio/blob/develop/agent/pkg/builder/container/types.go

clebs:
Hi @polaroi8d,
I find the project quite interesting and can write some unit tests for the agent.

clebs:
Hi @polaroi8d,
I tested the first pkg on the agent. I will test more packages when I have time but will always send PRs with one pkg each to not overwhelm the reviewers.

Looking forward to your feedback.

polaroi8d:
> Hi @polaroi8d, I tested the first pkg on the agent. I will test more packages when I have time but will always send PRs with one pkg each to not overwhelm the reviewers.
> 
> Looking forward to your feedback.

Good strategy, thanks! :)))

joremysh:
I could help with adding the unit tests. Is this issue still open?

polaroi8d:
@joremysh yea sure, thanks 🤚🏻

Dinoosawruss:
Happy to add some tests if still needed?

polaroi8d:
@Dinoosawruss go ahead!

Fall1ngStar:
👋 I'll add some tests !

polaroi8d:
@Fall1ngStar awesome!

SilverTux:
Hi,

I've created some tests for crane/k8s/error.go as I saw nobody worked on that part before. If there is something wrong with it please let me know and I'll update the PR accordingly.

joremysh:
I'm going to check with `agent/internal/crypt` first.

clebs:
I am testing `agent/internal/mapper` next!

clebs:
Hi @nandor-magyar I added a PR for the mapper that was previously on the agent. I saw it moved to `golang/internal` while I was implementing the tests.

I hope it is still fine.

polaroi8d:
[...]

PR #246, opened 2022-10-07
refactor(test): add unit tests for agent/api

clebs:
- Add unit tests for the agent/api pkg.
- Fixed wrong argument order in JSON unit test.
- `util.RemoveJSONComment` now also removes trailing spaces and tabs from comments.

Part of #241

### DIFF
--- a/agent/api/v1/deploy_test.go
+++ b/agent/api/v1/deploy_test.go
@@ -4,12 +4,14 @@
 package v1_test
 
 import (
+	"encoding/base64"
 	"testing"
 
 	"github.com/stretchr/testify/assert"
 
 	v1 "github.com/dyrector-io/dyrectorio/agent/api/v1"
 	"github.com/dyrector-io/dyrectorio/agent/internal/config"
+	"github.com/dyrector-io/dyrectorio/agent/pkg/builder/container"
 )
 
 func TestSetDeploymentDefaults(t *testing.T) {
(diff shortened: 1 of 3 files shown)


agent/internal/util/json.go
```diff
@@ -7,7 +7,7 @@ import (
 func RemoveJSONComment(strIn []byte) []byte {
 	str := string(strIn)
 	// edgecase that isn't handled: double quote in the comment
-	re1 := regexp.MustCompile(`//\s*?[^"\\S]*?\s*?(\n|\r|\r\n|\n\r)`)
+	re1 := regexp.MustCompile(`([ \t]*)//\s*?[^"\\S]*?\s*?(\n|\r|\r\n|\n\r)`)
```
nandor-magyar, line 10: \s includes tabs too.

> That is: \s matches a space, a tab, a carriage return, a line feed, or a form feed.
Source: https://www.regular-expressions.info/shorthand.html

Also, golang official: https://pkg.go.dev/regexp/syntax
```
\s             whitespace (== [\t\n\f\r ])
```

Is it necessary?

nandor-magyar (commented):
Nice catch of the argument order in tests. :+1:

agent/internal/util/json.go
clebs: You are right, `\s` is enough. Corrected the regex.

agent/api/v1/deploy_test.go
```diff
@@ -27,3 +29,202 @@ func TestSetDeploymentDefaults(t *testing.T) {
...
+}
+
+func TestUnmarshalJSON(t *testing.T) {
+	j := v1.Base64JSONBytes{}
+
+	err := j.UnmarshalJSON([]byte("this is not base64"))
+	assert.Error(t, err)
+
+	b := `
+{
```
polaroi8d, line 215: I think you missed formatting here.

polaroi8d (changes requested):
Please @clebs read the [CONTRIBUTING.md](https://github.com/dyrector-io/dyrectorio/blob/develop/CONTRIBUTING.md) we have a strict branching name convention (using eg. `refactor(test): add unit tests for agent` for PR title)

clebs:
Hi @polaroi8d,
thanks for the information, I have opened a new PR following the branching and title conventions here:
#257

PR #254, opened 2022-10-10
fix: prisma segfault

m8vago:
Fixes a segfault within the crux container image, which occurred with ssl based db connections.


robot9706 (approved):
LGTM 🚀

polaroi8d (approved):

PR #257, opened 2022-10-10
refactor(test): add unit tests for agent/api

clebs:
- Add unit tests for the agent/api pkg.
- Fixed wrong argument order in JSON unit test.
- util.RemoveJSONComment now also removes trailing spaces and tabs.

Part of #241

### DIFF
--- a/agent/internal/util/json.go
+++ b/agent/internal/util/json.go
@@ -7,7 +7,7 @@ import (
 func RemoveJSONComment(strIn []byte) []byte {
 	str := string(strIn)
 	// edgecase that isn't handled: double quote in the comment
-	re1 := regexp.MustCompile(`([ \t]*)//\s*?[^"\\S]*?\s*?(\n|\r|\r\n|\n\r)`)
+	re1 := regexp.MustCompile(`([\s]*)//\s*?[^"\\S]*?\s*?(\n|\r|\r\n|\n\r)`)
 	str = re1.ReplaceAllString(str, "")
 
 	return []byte(str)
--- a/agent/internal/util/json.go
+++ b/agent/internal/util/json.go
@@ -7,7 +7,7 @@ import (
 func RemoveJSONComment(strIn []byte) []byte {
 	str := string(strIn)
 	// edgecase that isn't handled: double quote in the comment
-	re1 := regexp.MustCompile(`//\s*?[^"\\S]*?\s*?(\n|\r|\r\n|\n\r)`)
+	re1 := regexp.MustCompile(`([ \t]*)//\s*?[^"\\S]*?\s*?(\n|\r|\r\n|\n\r)`)
 	str = re1.ReplaceAllString(str, "")
 
 	return []byte(str)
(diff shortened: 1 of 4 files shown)


agent/internal/util/json_test.go
```diff
@@ -18,8 +18,8 @@ func TestRemoveJSONComment(t *testing.T) {
 			  "i":"o",//
 "kek": "// //"//
 }`
-	expectstr := `{				            	    	  			  "json" : true, 			  "i":"o","kek": "// //"}`
+	expectstr := `{			  "json" : true,			  "i":"o","kek": "// //"}`
```
nandor-magyar, line 21: Test case does not cover "keeping new lines".

agent/internal/util/json.go
```diff
@@ -7,7 +7,7 @@ import (
 func RemoveJSONComment(strIn []byte) []byte {
 	str := string(strIn)
 	// edgecase that isn't handled: double quote in the comment
-	re1 := regexp.MustCompile(`//\s*?[^"\\S]*?\s*?(\n|\r|\r\n|\n\r)`)
+	re1 := regexp.MustCompile(`([^\S\r\n]*)//\s*?[^"\\S]*?\s*?(\n|\r|\r\n|\n\r)`)
```
nandor-magyar, line 10: Regexes are generally difficult to read, I don't see the added value, could you please explain?
`\S = ^\s` so this at the beginning is a double negation `^\S = ^^\s = \s`?

agent/internal/util/json_test.go
clebs: True, the test case I introduced on `deply_test.go` does though.

agent/internal/util/json.go
clebs: Of course, this group means, as you mention, double negate whitespace + negate newline chars. I am sure it could be done differently, but this in the end means "strip all whitespace characters except newlines".

I did try `\s` and what was happening is that new lines were stripped before the comment (not just at the end). This would result in lines being put all together in one.

nandor-magyar (approved):
Good job! :+1:

agent/internal/util/json_test.go
robot9706, line 21: "example" or "test" would be a better text instead of "kek" :D

robot9706 (changes requested):

agent/internal/util/json_test.go
clebs: I did not write this test, just adjusted it after improving the regex, `kek` was already there.

robot9706: Oh okay, nvm then :)

robot9706 (approved):

nandor-magyar:
There are missing commit signatures.

clebs:
@nandor-magyar I signed all commits now and should be passing all validations.

clebs:
Thanks for the quick merge! Also thank you for guiding me through the process.
Next PR will be easier 😉

PR #284, opened 2022-10-20
fix(cli): named pipe "deprecation", add socket to configuration

c3ppc3pp:
This PR adds support to older docker for windows installations, since newer ones use WSL2; and Docker SDK still using the old method so now the Traefik's socket is configurable, and if we stay at the default and we have a DOCKER_HOST env variable in use, we prefer that.


polaroi8d (approved):
LGTM (informal), but before merging please test it with some of our windows guys.

PR #288, opened 2022-10-21
fix(cli): traefik custom port

c3ppc3pp:
Kratos and CruxUI connection was faulty also Traefik were exposing the wrong port when we didn't used the default. Big thanks @robot9706 for pointing out this bug!


nandor-magyar (approved):

robot9706 (approved):
LGTM 📜

PR #291, opened 2022-10-21
feat(agent): silence containerbuilder's WithoutConflict func, rework WithoutConflict function, extend deleteContainer function

c3ppc3pp:
WithoutConflict builder function is reworked, only sets a flag, real work will happen in Create function
deleteContainer internal function now wont print errors unless it's necessary: detects running/exited containers and acts accordingly.

### DIFF
--- a/golang/pkg/builder/container/container_builder.go
+++ b/golang/pkg/builder/container/container_builder.go
@@ -72,6 +72,7 @@ type DockerContainerBuilder struct {
 	networks        []string
 	registryAuth    string
 	remove          bool
+	withoutConflict bool
 	restartPolicy   RestartPolicyName
 	entrypoint      []string
 	cmd             []string
@@ -219,9 +220,8 @@ func (dc *DockerContainerBuilder) WithTTY(tty bool) *DockerContainerBuilder {
 
 // Deletes the container with the given name if already exists.
 func (dc *DockerContainerBuilder) WithoutConflict() *DockerContainerBuilder {
-	if err := deleteContainer(dc.ctx, dc.containerName); err != nil {
-		log.Printf("builder could not stop/remove container (%s) to avoid conflicts: %s", dc.containerName, err.Error())
-	}
+	dc.withoutConflict = true
+
 	return dc
 }
 
@@ -280,6 +280,13 @@ func (dc *DockerContainerBuilder) Create() *DockerContainerBuilder {
 	if err := prepareImage(dc); err != nil {
 		logWrite(dc, fmt.Sprintf("Failed to prepare image: %s", err.Error()))
 		return dc
+	}
+
+	if dc.withoutConflict {
+		err := deleteContainer(dc.ctx, dc.containerName)
+		if err != nil {
+			logWrite(dc, fmt.Sprintf("Failed to resolve conflict during creating the container: %v", err))
+		}
 	}
 
 	portListNat := portListToNatBinding(dc.portRanges, dc.portList)
--- a/golang/pkg/builder/container/utils.go
+++ b/golang/pkg/builder/container/utils.go
@@ -138,15 +138,31 @@ func pullImage(ctx context.Context, logger io.StringWriter, fullyQualifiedImageN
 }
 
 func deleteContainer(ctx context.Context, containerName string) error {
-	if err := stopContainer(ctx, containerName); err != nil {
-		return err
+	deletableContainer, err := getContainer(ctx, containerName)
+	if err != nil {
+		return fmt.Errorf("builder could not get container (%s) to remove: %s", containerName, err.Error())
 	}
 
-	if err := removeContainer(ctx, containerName); err != nil {
-		return err
+	switch deletableContainer.State {
+	case "running", "paused", "restarting":
+		if err = stopContainer(ctx, containerName); err != nil {
+			return fmt.Errorf("builder could not stop container (%s): %s", containerName, err.Error())
+		}
+		fallthrough
+	case "exited", "dead", "created":
+		if err = removeContainer(ctx, containerName); err != nil {
+			return fmt.Errorf("builder could not remove container (%s): %s", containerName, err.Error())
+		}
+		return nil
+	case "":
+		// when there's no container we just skip it
+		return nil
+	default:
+		return fmt.Errorf("builder could not determine the state (%s) of the container (%s) for deletion: %s",
+			deletableContainer.State,
+			containerName,
+			err.Error())
 	}
-
-	return nil
 }
 
 func stopContainer(ctx context.Context, containerName string) error {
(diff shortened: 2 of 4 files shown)


golang/pkg/builder/container/container_builder.go
```diff
@@ -219,9 +219,30 @@ func (dc *DockerContainerBuilder) WithTTY(tty bool) *DockerContainerBuilder {
 
 // Deletes the container with the given name if already exists.
 func (dc *DockerContainerBuilder) WithoutConflict() *DockerContainerBuilder {
-	if err := deleteContainer(dc.ctx, dc.containerName); err != nil {
+	containers, err := getContainer(dc.ctx, dc.containerName)
```
nandor-magyar, line 222: Why the plural, it is just one container.

golang/pkg/builder/container/utils.go
```diff
@@ -176,6 +176,33 @@ func removeContainer(ctx context.Context, containerName string) error {
...
+
+	switch len(containers) {
+	case 0:
+		// when we didn't matched any
+		return types.Container{}, nil
+	case 1:
+		return containers[0], nil
+	default:
+		// shouldn't happen
+		return types.Container{}, fmt.Errorf("invalid match")
```
nandor-magyar, line 202: \* cannot happen, modify the error message to clearly communicate this, something like this: impossible, unreachable error, exact matching failed.

golang/pkg/cli/config_file.go
```diff
@@ -301,6 +301,14 @@ func SaveSettings(settings *Settings) {
 
 		// If settingspath is default, we create the directory for it
 		if settings.SettingsFilePath == settingspath {
+			if _, err := os.Stat(userconfdir); errors.Is(err, os.ErrNotExist) {
```
nandor-magyar, line 304: why is useconfdir variable not camelCased? :dagger:

nandor-magyar (changes requested):

golang/pkg/builder/container/container_builder.go
c3ppc3pp: true, but container would cause name-collision with an imported dependency

c3ppc3pp:
@nandor-magyar , as per our conversation I'll wait for the zerolog PR, then I'll refactor this quite a bit.

golang/pkg/builder/container/utils.go
c3ppc3pp: last one sounds nice to me, thanks!

golang/pkg/cli/config_file.go
c3ppc3pp: changed in every instance

golang/pkg/builder/container/container_builder.go
c3ppc3pp: fixed

golang/pkg/builder/container/utils.go
```diff
@@ -176,6 +192,33 @@ func removeContainer(ctx context.Context, containerName string) error {
 	return nil
 }
 
+// Check the existence of a container, then return it, matches only one
+func getContainer(ctx context.Context, containerName string) (types.Container, error) {
```
polaroi8d, line 196: Can't we use the `GetContainersByName` from dagent utils?

golang/pkg/builder/container/utils.go
```diff
@@ -138,15 +138,31 @@ func pullImage(ctx context.Context, logger io.StringWriter, fullyQualifiedImageN
...
+	switch deletableContainer.State {
+	case "":
+		// when there's no container we just skip it
+		return nil
+	case "running":
+		if err = stopContainer(ctx, containerName); err != nil {
+			return fmt.Errorf("builder could not stop container (%s): %s", containerName, err.Error())
+		}
+		fallthrough
+	case "exited":
```
polaroi8d, line 155: What about the other states? For example, if a container is `dead` or `paused`? I think we have to implement all states.

golang/pkg/builder/container/utils.go
```diff
@@ -138,15 +138,31 @@ func pullImage(ctx context.Context, logger io.StringWriter, fullyQualifiedImageN
...
-		return err
+	deletableContainer, err := getContainer(ctx, containerName)
+	if err != nil {
+		return fmt.Errorf("builder could not get container (%s) to remove: %s", containerName, err.Error())
 	}
 
-	if err := removeContainer(ctx, containerName); err != nil {
-		return err
+	switch deletableContainer.State {
+	case "":
```
polaroi8d, line 147: Putting this before default would be better, IMHO.

polaroi8d (changes requested):

golang/pkg/builder/container/utils.go
c3ppc3pp: Not really since it's filter is not adequate sadly.

c3ppc3pp:
```
	containers, err := cli.ContainerList(ctx, types.ContainerListOptions{
		All:     true,
		Filters: filters.NewArgs(filters.KeyValuePair{Key: "name", Value: fmt.Sprintf("^/?%s-", name)}),
	})
```
name filter could return more than one container, and it filters by a prefix judging by looks

golang/pkg/builder/container/utils.go
c3ppc3pp: Oh, yeah, and restarting.... good catch!

polaroi8d: Or created 🪄

golang/pkg/builder/container/utils.go
c3ppc3pp: fixed

golang/pkg/builder/container/utils.go
c3ppc3pp: fixed :sparkles:

golang/pkg/builder/container/utils.go
polaroi8d: I would prefer to make smarter the first function and reuse it since this is a code duplication of the function doing the same thing, what do you think @nandor-magyar?

nandor-magyar: Yes we could make it happen to avoid code duplication, the function needs  a boolean to toggle exact matching, in such case would prepend/append the regex parts if missing.
There an other thing troubling me here, that is builder using dagent code, not vice-versa -- that would be the correct way. All the docker functions we use could go into an isolated docker package since both of the use is aggressively.

polaroi8d: An approved way for me.

c3ppc3pp: We talked about it with @nandor-magyar , this is a much needed change, however it's a different scope, and will be implemented in different PR too.

polaroi8d: I don't like this approach, we'll forget it. If @nandor-magyar or you create a ticket for this, I can approve it.

polaroi8d: Or do it instantly after this PR 😄

c3ppc3pp: Made a ticket, will do it later this week. (I wanna finish my current task first)

polaroi8d: Thanks 🙏🏻

nandor-magyar (approved):
LGTM

Issue #217, opened 2022-09-29
Create UUIDv4 validation for the NodeID config string


polaroi8d:
In `agent/internal/config/config.go` we store our configuration defaults. In `line 28` you can find NodeID which is an UUIDv4 string, but we don't validate it. Please create a validation for this particular data.

If you are ready with the task, please remove the `// TODO` line from the file. 

More info: 
A universally unique identifier (`UUID`) is an identifier standard used in software construction. The meaning of each bit is defined by any of several variants. It is a 128-bit value. Currently, UUID's are as specified in RFC4122. [https://www.rfc-editor.org/rfc/rfc4122#section-4.1.7](https://www.rfc-editor.org/rfc/rfc4122#section-4.1.7)

Akshu-on-github:
Hello there, I'd like to take this issue up

polaroi8d:
@raghav-rama for sure. If you have a question just drop us a message on [Discord](https://discord.gg/FHq3FMKt).

raghav-rama:
To open PR, first I need to be assigned, right?

polaroi8d:
@raghav-rama I am assigned to you.

raghav-rama:
![image](https://user-images.githubusercontent.com/91389059/197160656-87c63653-dbdb-4ef5-ad6f-36259e506842.png)
I've done this much, but I can't understand how you are taking the string as input. Can you explain that @polaroi8d?

nandor-magyar:
It is done by the library we use: cleanenv. You have to create a new type alias, `type UUID string`, then use setValue() function with correct types and parameters. Similar others in the config folder.
See lib docs and examples here too: https://github.com/ilyakaznacheev/cleanenv

PR #292, opened 2022-10-22
feat: uuidv4-validation-NodeID-config-string

raghav-rama:
Issue no. #217 
Please review, I'm not sure if this is the right way to do it.


golang/internal/config/config.go
```diff
@@ -25,10 +31,43 @@ type CommonConfiguration struct {
...
-	NodeID            string         `yaml:"nodeID" env:"NODE_ID" env-default:"cb7e9573-9a43-4d5b-8005-eb8bb7a423c4"`
+	NodeID            UUID           `yaml:"nodeID" env:"NODE_ID" env-default:"cb7e9573-9a43-4d5b-8005-eb8bb7a423c4"`
 	ReadHeaderTimeout time.Duration  `yaml:"readHeaderTimeout"    env:"READ_HEADER_TIMEOUT"      env-default:"15s"`
 	Registry          string         `yaml:"registry"             env:"REGISTRY"                 env-default:"index.docker.io"`
 	RegistryPassword  string         `yaml:"registryPassword"     env:"REGISTRY_PASSWORD"        env-default:""`
 	RegistryUsername  string         `yaml:"registryUsername"     env:"REGISTRY_USERNAME"        env-default:""`
 	SecretPrivateKey  ConfigFromFile `yaml:"secretPrivateKeyFile" env:"SECRET_PRIVATE_KEY_FILE"  env-default:"/srv/dagent/private.key"`
 }
+
+var cfg CommonConfiguration
```
nandor-magyar, line 42: commonConfig itself is never used, should not be used.

golang/internal/config/config.go
```diff
@@ -25,10 +31,43 @@ type CommonConfiguration struct {
...
+
+func ReadConfig() error {
```
nandor-magyar, line 44: This is unnecessary.

nandor-magyar (changes requested):
Not quite, you only need to define a type alias (that part is correct) and a simple function like it is done here in the example of the library in use:
https://github.com/ilyakaznacheev/cleanenv/blob/3bb055c0867b82fa8054e6ca82d64f1c30e06ea7/example/custom_setter/main.go#L37

In your case this might look like
```golang
func (u *UUID) SetValue(raw string) {
// validation

// if valid
*u = UUID(raw)
}
```

The SetValue is public so the parser can call it when it maps the configuration type, this works the same way as some other marshal/unmarshalers, json and yaml for example.

You could also add some unit test into a config_test.go file, that might help you as well.

golang/internal/config/config.go
raghav-rama: Can u pls tell why?

raghav-rama:
I'm on it🚀

polaroi8d:
> I'm on it🚀

Awesome, thank you. Please do not miss the commit signature.

golang/internal/config/config.go
raghav-rama: Oooh, I see.

golang/internal/config/config.go
```diff
@@ -19,16 +25,23 @@ type CommonConfiguration struct {
...
-	NodeID            string         `yaml:"nodeID" env:"NODE_ID" env-default:"cb7e9573-9a43-4d5b-8005-eb8bb7a423c4"`
-	ReadHeaderTimeout time.Duration  `yaml:"readHeaderTimeout"    env:"READ_HEADER_TIMEOUT"      env-default:"15s"`
-	Registry          string         `yaml:"registry"             env:"REGISTRY"                 env-default:"index.docker.io"`
-	RegistryPassword  string         `yaml:"registryPassword"     env:"REGISTRY_PASSWORD"        env-default:""`
-	RegistryUsername  string         `yaml:"registryUsername"     env:"REGISTRY_USERNAME"        env-default:""`
-	SecretPrivateKey  ConfigFromFile `yaml:"secretPrivateKeyFile" env:"SECRET_PRIVATE_KEY_FILE"  env-default:"/srv/dagent/private.key"`
-	// GRPC token is set separately, because nested structures are not yet suppported in cleanenv
-	GrpcToken *ValidJWT
+	// TODO(c3ppc3pp): custom setter to validate input
+	GrpcToken            string         `yaml:"grpcToken"             env:"GRPC_TOKEN"`
```
nandor-magyar, line 29: It seems that you have the wrong set of changes in your changes, the deleted ones are the valid ones to keep except the custom type of UUID, but make sure to remove the UUIDv4 comment.

nandor-magyar (changes requested):
Something might went wrong while merging, see my comment.
Make sure to run go mod tidy before committing.

golang/internal/config/config.go
raghav-rama: oops, my bad. I'll fix it in a jiffy

raghav-rama: > except the custom type of UUID

Excuse me for not understanding, but should I also remove the custom UUID type?
Because I have used it for SetValue()...

raghav-rama:
If it still doesn't work, I'll close this PR and open a fresh new one.

golang/internal/config/config.go
nandor-magyar: But the order is still messed up, you should not have anything else changed, but the custom type, the new type is good as it is right now.

raghav-rama: Okay, GrpcToken should be in the end, right?

Also, will it affect behavior?

nandor-magyar: > Okay, GrpcToken should be in the end, right?
Yes.
> Also, will it affect behavior?
The behavior is not changed, the git history is cleaner and better representing what is going on.

raghav-rama: > Yes.

Done😀

polaroi8d:
Closed and reopened in #328 // Added @raghav-rama as a co-author.

PR #294, opened 2022-10-24
fix(ci): missing PR trigger in golang.yml, removed duplicated triggers

c3ppc3pp:


balanceee (approved):

.github/workflows/golang.yml
```diff
@@ -1,6 +1,13 @@
 name: golang
 on:
   push:
+    branches:
+      - develop
```
nandor-magyar, line 5: This allows other branches to use the workflows because of the `pull_request` section below, right?

c3ppc3pp: Yes, and we wont have duplicated pipelines under a PR because of this, and we wont run these for every commit on every branch just when the given branch is near finished(PR submitted).

nandor-magyar (approved):
NIce catch mon!

PR #295, opened 2022-10-24
fix(agent): remove comment and version strings from pgp keys

c3ppc3pp:
default custom headers in PGP blocks are not deterministic because their order is not fixed

so I remove them in this PR


nandor-magyar (approved):
Nice! Rebase please! :+1:

polaroi8d (approved):

PR #298, opened 2022-10-25
chore: actualize vscode debug paths

nandor-magyar:
I've noticed these remained old, I might've stashed these instead of committing.

### DIFF
--- a/.vscode/launch.json
+++ b/.vscode/launch.json
@@ -25,8 +25,8 @@
             "type": "go",
             "request": "launch",
             "mode": "auto",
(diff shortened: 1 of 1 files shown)


polaroi8d (approved):
More like a fix instead of a clean. (We don't have a clean topic) Otherhand LGTM but please rename the PR title.

robot9706 (approved):
LGTM 👀

c3ppc3pp (approved):

polaroi8d:
> @polaroi8d I wanted chore, but messed up, chore because it doesn't effect code in any way.

But fix, because right now lunching application with vscode doesn't work.

PR #299, opened 2022-10-25
feat(agent): rework installscript to respect rootless/podman or other…

c3ppc3pp:
…wise different hosts; expose mailslurper smtpport

when I started I came top a realization: i need to expose SMTP as well (nobody told me which means you did not used it? :smiling_imp: )

main scope is reworking dagent install script so it will work in wider array of environments (podman, doesn't need root, etc) but these need set environment variables which could be added to the UI later


nandor-magyar (approved):
Looks good, the only thing missing is a new line at the end of the script file.

c3ppc3pp:
> Looks good, the only thing missing is a new line at the end of the script file.

There is one tho. No EOL is marked with a special icon in GitHub.

polaroi8d (approved):

PR #305, opened 2022-10-28
feat: merge go package into single module - go install

nandor-magyar:
This is started as a little experiment, but as I've tested it, this works without any modification to local envs or pipelines.

Notes:
- binaries won't have different size
- no changes to paths
- no changes to Makefiles

This will probably let us `go install` cli (or crane or dagent).


polaroi8d (approved):
Looks okay for me.

c3ppc3pp (approved):
Looks aight.

PR #310, opened 2022-10-30
Changes Version Format Regex

Swap-Nova:
I changed a little bit by adding ```-``` in the version_test.go file.


package.json
```diff
@@ -0,0 +1,22 @@
+{
```
polaroi8d, line 1: I think we do not need package.jsons in the root folder.

polaroi8d:
Closing the PR due the inactivity.

PR #313, opened 2022-11-02
refactor(dagent): common docker functions

c3ppc3pp:


golang/internal/dockerhelper/container.go
```diff
@@ -0,0 +1,109 @@
+package dockerhelper
```
polaroi8d, line 1: The naming is weird for me, using camelCase or snake_case would be better IMHO. The official documentation says these 2 can be okay for naming the packages. Link: https://go.dev/blog/package-names

I will vote for camelCase, but wait for @nandor-magyar and @robot9706 to agree on this.

golang/internal/dockerhelper/container.go
```diff
@@ -0,0 +1,109 @@
...
+}
+
+// Using exact match!
+func GetOneContainerByName(ctx context.Context, dog *dogger.DeploymentLogger, nameFilter string) (types.Container, error) {
+	containers, err := GetContainersByName(ctx, exactMatch(nameFilter))
+	if err != nil {
+		return types.Container{}, err
+	}
+
+	switch len(containers) {
```
polaroi8d, line 97: `ContainerListOptions` there is a Limit int attribute, can't we workaround this with the limit?

golang/internal/dockerhelper/dockerhelper_test.go
```diff
@@ -0,0 +1,11 @@
+package dockerhelper
+
+import "context"
+
+func TestDeleteContainer(ctx context.Context, container string) error {
+	return DeleteContainerByName(ctx, nil, container)
+}
+
+func TestDeleteNetwork(ctx context.Context, networkID string) error {
+	return DeleteNetworkByID(ctx, networkID)
```
polaroi8d, line 10: Missing Given, Then :D

golang/pkg/dagent/utils/docker.go
```diff
@@ -336,7 +336,8 @@ func DeployImage(ctx context.Context,
 	// err is ignored because it means no container is available
 	state, _ := CheckContainer(deployImageRequest.RequestID, containerName)
 
-	err = checkContainerState(dog, containerName, state)
+	// TODO : delete
```
polaroi8d, line 339: TODO comment.

golang/internal/dockerhelper/container.go
```diff
@@ -0,0 +1,109 @@
...
+
+	"github.com/docker/docker/api/types"
+	"github.com/docker/docker/api/types/filters"
+	"github.com/docker/docker/client"
+	"github.com/dyrector-io/dyrectorio/golang/internal/dogger"
+)
+
+const dockerClientTimeoutSeconds = 30
+
+func DeleteContainerByName(ctx context.Context, dog *dogger.DeploymentLogger, nameFilter string) error {
```
polaroi8d, line 16: No usage of  `dog *dogger.DeploymentLogger` in the function.

golang/internal/dockerhelper/container.go
c3ppc3pp: Name changed for a more hierarchical pattern, similar to containerbuilder.

golang/internal/dockerhelper/container.go
```diff
@@ -0,0 +1,109 @@
...
+	containers, err := cli.ContainerList(ctx, types.ContainerListOptions{All: true, Filters: filter})
+	if err != nil {
+		return []types.Container{}, err
+	}
+
+	return containers, err
+}
+
+// Using exact match!
+func GetOneContainerByName(ctx context.Context, dog *dogger.DeploymentLogger, nameFilter string) (types.Container, error) {
```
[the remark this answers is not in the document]
c3ppc3pp: fixed, thanks!

golang/internal/dockerhelper/container.go
c3ppc3pp: Thing is when we do a "find me THAT one container" type of operation we shouldn't use limits. Issue is if the search filter isn't specific enough we could end up with more than one container, and we making space for doing an unwanted operation on a container we do not care about. I'd rather rule out the chance or opportunity to work on a wrong container.

golang/internal/dockerhelper/dockerhelper_test.go
c3ppc3pp: This will be reworked.

golang/pkg/dagent/utils/docker.go
c3ppc3pp: good catch :see_no_evil:

golang/internal/dockerhelper/container.go
c3ppc3pp: dogger wasnt implemented in this draft yet.

golang/internal/helper/docker/network.go
```diff
@@ -0,0 +1,66 @@
...
+		Filters: filter,
+	}
+
+	networks, err := cli.NetworkList(ctx, networkListOption)
+	if err != nil {
+		return fmt.Errorf("error list existing networks: %w", err)
+	}
+
+	if len(networks) > 0 {
+		log.Error().Msg("Provided network name: " + name + " exists. Skip to create new network.")
```
polaroi8d, line 41: Use `Str(key, value)` with `Msg()`

golang/internal/helper/docker/network.go
```diff
@@ -0,0 +1,66 @@
...
+	"fmt"
+
+	"github.com/docker/docker/api/types"
+	"github.com/docker/docker/api/types/filters"
+	"github.com/docker/docker/client"
+	"github.com/rs/zerolog/log"
+)
+
+func GetAllNetworks() ([]types.NetworkResource, error) {
+	ctx := context.Background()
```
polaroi8d, line 14: We get `ctx := context.Background()` from parameters in every function. Using the same way would be better.

golang/internal/helper/docker/network.go
```diff
@@ -0,0 +1,66 @@
...
+	return nil
+}
+
+func DeleteNetworkByID(ctx context.Context, networkID string) error {
+	cli, err := client.NewClientWithOpts(client.FromEnv, client.WithAPIVersionNegotiation())
+	if err != nil {
+		panic(err)
+	}
+
+	return cli.NetworkRemove(ctx, networkID)
```
polaroi8d, line 65: Don't we need to check errors after removing networks?

golang/pkg/builder/container/container_builder.go
```diff
@@ -16,6 +16,8 @@ import (
 	"github.com/docker/docker/client"
 	"github.com/docker/go-connections/nat"
 	"github.com/rs/zerolog/log"
+
+	dockerhelper "github.com/dyrector-io/dyrectorio/golang/internal/helper/docker"
```
polaroi8d, line 20: `dockerHelper`?

polaroi8d (changes requested):

golang/internal/helper/docker/network.go
c3ppc3pp: We do need it, but in different place, where this func is called. (and as you might rightfully assume: there was none :D )

golang/pkg/builder/container/container_builder.go
c3ppc3pp: Riiight. Fixed. Also fixed imagehelper -> imageHelper

golang/internal/helper/docker/network.go
c3ppc3pp: fixed

golang/internal/helper/docker/container.go
```diff
@@ -0,0 +1,130 @@
...
+
+func DeleteContainerByName(ctx context.Context, dog *dogger.DeploymentLogger, nameFilter string) error {
+	matchedContainer, err := GetContainerByName(ctx, nil, nameFilter)
+	if err != nil {
+		return fmt.Errorf("builder could not get container (%s) to remove: %s", nameFilter, err.Error())
+	}
+
+	switch matchedContainer.State {
+	case "running", "paused", "restarting":
+		if err = StopContainerByName(ctx, dog, nameFilter); err != nil {
```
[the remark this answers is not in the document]
c3ppc3pp: This pattern was taken from the original func, but fixed nonetheless

golang/internal/helper/docker/network.go
c3ppc3pp: Also fixed.

golang/internal/helper/docker/container.go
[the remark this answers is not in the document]
polaroi8d: Okey, thanks.

golang/internal/dockerhelper/container.go
c3ppc3pp: @polaroi8d what do you think about this issue?

polaroi8d: Got it, but what if the search filter isn't specific enough, what guarantees that I need the first container in the list? Because you get the first element from the list, didn't see this is enough. 

Code snippet:
```
	case 1:
		return containers[0], nil
```

c3ppc3pp: That's the idea, it's another safety to search for every matching container, and if we have more than one, we return error.
```
	switch len(containers) {
	// Ideal, where we have just one matching
	case 1:
		return containers[0], nil
	// Ideal, where we have no match
	case 0:
		return types.Container{}, nil
	// Not ideal at all, when the search pattern matched multiple, so it wasn't an exact match, this should not happen, but I'd like to keep it for safety
	default:
		return types.Container{}, fmt.Errorf("more than one matching container")
	}
```

When the search input is "random-container" and we return random-container, random-container2, random-container3 etc; this shouldn't happen because we use exact matching pattern (which isn't the default somehow...) earlier in the func.

MFW when docker does not use exact matching by default: https://static.guim.co.uk/sys-images/Guardian/Pix/pictures/2014/10/30/1414670814763/1956-MOBY-DICK-012.jpg

By all means, this code isn't that necessary but an extra safety measure.

polaroi8d: Ahh okay, thanks for the clarification.

golang/internal/helper/docker/container.go
```diff
@@ -0,0 +1,132 @@
...
+	}
+	if err := cli.ContainerStop(ctx, nameFilter, &timeoutValue); err != nil {
+		return err
+	}
+
+	return nil
+}
+
+// Matches one
+func RemoveContainerByName(ctx context.Context, dog *dogger.DeploymentLogger, nameFilter string) error {
```
balanceee, line 66: Why is it public?

c3ppc3pp: This is a helper lib. Functions have to be public, there is chance we do not use all of them currently but I'd like to keep the opportunity open.

balanceee: Yeah but in this case we should only use the DeleteContainerByName function, or am i missing something?

c3ppc3pp: We only use that one because most of the time we do "I don't care i just want to see this container GONE" and this is what that function does. 

If we would want to stop a running or remove a stopped container we would have to explicitly use these. These aren't in our scope right now, but this might change in the future.

golang/internal/helper/docker/container_test.go
```diff
@@ -0,0 +1,185 @@
...
+	containerNames []string
+}
+
+const (
+	nginxImage   = "docker.io/library/nginx:latest"
+	prefixLength = 8
+)
+
+// Runs before everything
+func (testsuite *DockerContainerHelperTestSuite) SetupSuite() {
```
polaroi8d, line 39: Naming these BeforeAll or something like that, but @szszigeti can help with the naming convention, IMHO.

golang/internal/helper/docker/container_test.go
```diff
@@ -0,0 +1,185 @@
...
+	}
+}
+
+// Runs after everything
+func (testsuite *DockerContainerHelperTestSuite) TearDownSuite() {
+	testsuite.dockerClient.Close()
+}
+
+// this function executes before each test case
+func (testsuite *DockerContainerHelperTestSuite) SetupTest() {
```
polaroi8d, line 64: `BeforeAll()`

golang/internal/helper/docker/container_test.go
```diff
@@ -0,0 +1,185 @@
...
+		log.Fatal().Msg("Could not start container")
+	}
+	_, err = testsuite.testContainers[2].Create().Start()
+	if err != nil {
+		log.Fatal().Msg("Could not start container")
+	}
+}
+
+// this function executes after each test case
+func (testsuite *DockerContainerHelperTestSuite) TearDownTest() {
```
polaroi8d, line 80: `AfterAll()`

golang/internal/helper/docker/container_test.go
```diff
@@ -0,0 +1,185 @@
...
+		preparedContainer := containerbuilder.NewDockerBuilder(context.Background()).
+			WithImage(nginxImage).
+			WithName(fmt.Sprintf("%s-%s", testsuite.prefix, testsuite.containerNames[i])).
+			WithRestartPolicy(containerbuilder.NoRestartPolicy)
+		testsuite.testContainers = append(testsuite.testContainers, preparedContainer)
+	}
+}
+
+// Runs after everything
+func (testsuite *DockerContainerHelperTestSuite) TearDownSuite() {
```
polaroi8d, line 59: `After()`

golang/internal/helper/docker/container_test.go
```diff
@@ -0,0 +1,185 @@
...
+	assert.Nil(testsuite.T(), err, "should not have errors")
+
+	err = dockerHelper.StopContainerByName(testsuite.ctx, nil, contName)
+
+	assert.Nil(testsuite.T(), err, "should not have errors")
+	matched, err = dockerHelper.GetContainerByName(testsuite.ctx, nil, contName)
+
+	assert.Equal(testsuite.T(), "exited", matched.State,
+		"should return a stopped container")
+	assert.Nil(testsuite.T(), err, "should not have errors")
```
polaroi8d, line 152: If you want to check the err is nil, use `assert.noError()` function.

golang/internal/helper/docker/network_test.go
```diff
@@ -0,0 +1,116 @@
+//go:build unit
```
polaroi8d, line 1: Same comments here.

polaroi8d (approved):
Minor issues mentioned, lgtm after doing the changes.

golang/internal/helper/docker/container_test.go
c3ppc3pp: AFAIK, these function names have to be named like that. This is how this testing suite works. :shrug:

polaroi8d: Hmm, okey than sorry.

golang/internal/helper/docker/container_test.go
c3ppc3pp: Good catch! Thanks! Replaced those.

golang/internal/helper/docker/container_test.go
```diff
@@ -0,0 +1,182 @@
...
+}
+
+// Runs after everything
+func (testsuite *DockerContainerHelperTestSuite) TearDownSuite() {
+	testsuite.dockerClient.Close()
+}
+
+// this function executes before each test case
+func (testsuite *DockerContainerHelperTestSuite) SetupTest() {
+	_, err := testsuite.testContainers[0].Create().Start()
```
robot9706, line 65: A `for` loop would be better I think.

robot9706 (changes requested):

golang/internal/helper/docker/container_test.go
c3ppc3pp: Thanks, fixed!

golang/internal/helper/docker/network_test.go
```diff
@@ -0,0 +1,116 @@
...
+}
+
+// Runs after everything
+func (testsuite *DockerNetworkHelperTestSuite) TearDownSuite() {
+	testsuite.dockerClient.Close()
+}
+
+// this function executes before each test case
+func (testsuite *DockerNetworkHelperTestSuite) SetupTest() {
+	for i := range testsuite.networkNames {
```
nandor-magyar, line 49: for each testcase all the networks are created, why?

golang/pkg/dagent/dagent.go
```diff
@@ -39,7 +39,7 @@ func Serve(cfg *config.Configuration) {
 	grpc.Init(grpcContext, grpcParams, &cfg.CommonConfiguration, grpc.WorkerFunctions{
 		Deploy:     utils.DeployImage,
 		Watch:      utils.GetContainersByNameCrux,
-		Delete:     utils.DeleteContainerByName,
+		Delete:     utils.DeleteContainerByNameGrpc,
```
nandor-magyar, line 42: This suffix is weird, these are all grpc functions, could this be solved in another way?

golang/internal/helper/docker/network_test.go
c3ppc3pp: Consistency. So you can write test for manipulating those networks and expect those to be there by default.

golang/internal/helper/docker/container_test.go
```diff
@@ -0,0 +1,176 @@
...
+			log.Warn().Err(err).Msg("")
+		}
+	}
+}
+
+func (testsuite *DockerContainerHelperTestSuite) TestGetAllContainers() {
+	containers, err := dockerHelper.GetAllContainers(testsuite.ctx, nil)
+
+	assert.NoError(testsuite.T(), err, "%s", err)
+	assert.LessOrEqual(testsuite.T(), len(testsuite.testContainers), len(containers),
```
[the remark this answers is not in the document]
c3ppc3pp: If you use it locally there might be other containers, or even in the actions you cant be sure how many there will be. At least these containers, and plus one where the tests run in Github Pipelines. Other envs are more unreliable with this.

golang/internal/helper/docker/container_test.go
```diff
@@ -0,0 +1,176 @@
...
+		preparedContainer := containerbuilder.NewDockerBuilder(context.Background()).
+			WithImage(nginxImage).
+			WithName(fmt.Sprintf("%s-%s", testsuite.prefix, testsuite.containerNames[i])).
+			WithRestartPolicy(containerbuilder.NoRestartPolicy)
+		testsuite.testContainers = append(testsuite.testContainers, preparedContainer)
+	}
+}
+
+// Runs after everything
+func (testsuite *DockerContainerHelperTestSuite) TearDownSuite() {
```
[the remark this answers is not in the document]
c3ppc3pp: That would shadow the import dep. :)

golang/internal/helper/docker/container.go
```diff
@@ -0,0 +1,132 @@
...
+			matchedContainer.State,
+			nameFilter,
+			err.Error())
+	}
+}
+
+func StopContainerByName(ctx context.Context, dog *dogger.DeploymentLogger, nameFilter string) error {
+	cli, err := client.NewClientWithOpts(client.FromEnv, client.WithAPIVersionNegotiation())
+	if err != nil {
+		log.Fatal().Err(err).Msg("")
```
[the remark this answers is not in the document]
c3ppc3pp: fixed

golang/pkg/dagent/dagent.go
c3ppc3pp: We talked about this in person, fixed.

golang/internal/helper/docker/container_test.go
```diff
@@ -0,0 +1,176 @@
...
+
+func (testsuite *DockerContainerHelperTestSuite) TestGetAllContainers() {
+	containers, err := dockerHelper.GetAllContainers(testsuite.ctx, nil)
+
+	assert.NoError(testsuite.T(), err, "%s", err)
+	assert.LessOrEqual(testsuite.T(), len(testsuite.testContainers), len(containers),
+		"should return at least as many containers as we started/created")
+}
+
+func (testsuite *DockerContainerHelperTestSuite) TestGetAllContainersbyName() {
```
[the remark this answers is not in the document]
c3ppc3pp: Fixed.

golang/internal/helper/docker/container_test.go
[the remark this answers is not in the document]
c3ppc3pp: I made them camelCased.

golang/internal/helper/docker/container_test.go
```diff
@@ -0,0 +1,176 @@
...
+const (
+	nginxImage   = "docker.io/library/nginx:latest"
+	prefixLength = 8
+)
+
+// Runs before everything
+func (testsuite *DockerContainerHelperTestSuite) SetupSuite() {
+	cli, err := client.NewClientWithOpts(client.FromEnv, client.WithAPIVersionNegotiation())
+	if err != nil {
+		log.Fatal().Msg("Could not connect to docker socket.")
```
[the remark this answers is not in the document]
c3ppc3pp: Fixed. Good catch!

golang/pkg/builder/container/container_builder_test.go
```diff
@@ -14,18 +14,24 @@ import (
...
 	"github.com/stretchr/testify/assert"
 
+	dockerHelper "github.com/dyrector-io/dyrectorio/golang/internal/helper/docker"
 	containerbuilder "github.com/dyrector-io/dyrectorio/golang/pkg/builder/container"
 )
 
 func builderCleanup(builder *containerbuilder.DockerContainerBuilder) {
+	ctx := context.Background()
 	if builder.GetContainerID() != nil {
-		containerbuilder.DeleteContainer(context.Background(), *builder.GetContainerID())
```
robot9706, line 24: Talked in person, this should be 'delete by id'.

robot9706 (changes requested):

golang/internal/helper/docker/container_test.go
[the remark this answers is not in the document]
nandor-magyar: The right assertion then MoreOrEqual, no?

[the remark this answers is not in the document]
c3ppc3pp: We talked about it IRL, testing suite has some twisted logic.

robot9706 (approved):
LGTM 🚀

PR #314, opened 2022-11-03
feat(crux): add secrets to template

peterambrus3:
From now you can add your secrets to the template.

### DIFF
--- a/web/crux/src/services/template.file.service.ts
+++ b/web/crux/src/services/template.file.service.ts
@@ -3,7 +3,7 @@ import { readdir, readFileSync } from 'fs'
 import { join, parse } from 'path'
 import { cwd } from 'process'
 import { CreateRegistryRequest, Port, TemplateResponse, Volume } from 'src/grpc/protobuf/proto/crux'
-import { ContainerConfigData, UniqueKeyValue } from 'src/shared/model'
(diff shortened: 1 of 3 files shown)


polaroi8d (approved):
Looks good to me 👏🏻

web/crux/src/services/template.file.service.ts
```diff
@@ -32,6 +32,7 @@ type TemplateConfig = {
   volumes: TemplateVolume[]
   ports: Port[]
   environment: UniqueKeyValue[]
+  secrets: UniqueKey[]
```
balanceee, line 35: I think it should be UniqueSecretKey.

web/crux/src/services/template.file.service.ts
```diff
@@ -44,6 +45,7 @@ export interface TemplateImage {
   capabilities: UniqueKeyValue[]
   config: TemplateContainerConfig
   environment: UniqueKeyValue[]
+  secrets: UniqueKey[]
```
balanceee, line 48: Same as above

web/crux/src/services/template.file.service.ts
peterambrus3: true 🤦‍♂️

balanceee (approved):
LGTM

PR #326, opened 2022-11-10
fix(crux): remove cert related code from Dockerfile

nandor-magyar:
Pipeline is broken, here is the solution :cat2:


robot9706 (approved):
🚀

polaroi8d (approved):

PR #337, opened 2022-11-14
feat(crux): add Prometheus basics

polaroi8d:
Add Prometheus with an agent connection counter.

### DIFF
--- a/web/crux/src/app/agent/agent.service.ts
+++ b/web/crux/src/app/agent/agent.service.ts
@@ -2,6 +2,8 @@ import { Injectable, Logger } from '@nestjs/common'
 import { ConfigService } from '@nestjs/config'
 import { JwtService } from '@nestjs/jwt'
 import { DeploymentEventTypeEnum, DeploymentStatusEnum, NodeTypeEnum } from '@prisma/client'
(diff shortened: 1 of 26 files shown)


web/crux/src/app/agent/agent.service.ts
```diff
@@ -302,6 +305,7 @@ export default class AgentService {
     connection.status().subscribe(it => this.onAgentConnectionStatusChange(agent, it))
 
     this.logger.log(`Agent joined with id: ${request.id}, key: ${!!agent.publicKey}`)
+    this.agent_counter.inc()
```
robot9706, line 308: Does this decrease when an agent disconnects?

robot9706 (changes requested):

web/crux/src/app/agent/agent.service.ts
polaroi8d: Nope, this is a counter function that represents a single increasing counter whose value can only increase or be reset to zero on restart.

robot9706 (approved):
LGTM, Tested, metrics endpoint seems to be working properly 🚀

peterambrus3 (approved):
lgtm, tested locally

PR #344, opened 2022-11-16
fix(crux): respect prefix after successful deployment

balanceee:
Previously every deployment became obsolete on the version. Now only those which has same prefix as the new deployment.


balanceee:
Stays draft until it can be testable (can't deploy at the moment).

m8vago (approved):

robot9706 (approved):
LGTM 🦔

PR #347, opened 2022-11-16
fix(crux): increase version mapper fix

peterambrus3:


polaroi8d (approved):
LGTM

balanceee (changes requested):
Code LGTM and tested locally 🆗. But I still can not add a new version to a complex product after it has a version already. I know its not part of the PR but the problem is exatly the same -> config mapping error. Can you please also fix this in your PR?

balanceee (approved):
LGTM & Tested

PR #349, opened 2022-11-16
fix(crux-ui): registry field validator

bvektor:


web/crux-ui/src/validations/registry.ts
```diff
@@ -16,7 +16,7 @@ const shouldResetMetaData = { reset: true }
 const registryCredentialRole = yup
   .mixed()
   .meta(shouldResetMetaData)
-  .when(['type', '_private'], {
+  .when(['type', 'private'], {
```
robot9706, line 19: imo you should also fix the argument few lines below.

Also `registrySchema ` contains a `_private` shouldn't that be fixed too?

robot9706 (changes requested):

web/crux-ui/src/validations/registry.ts
bvektor: 'private' is a reserved word in strict mode. Everywhere else we can use it without '_'

robot9706 (approved):

m8vago (approved):

PR #351, opened 2022-11-16
fix(agent): dc.logger nil check fix, add option to ignore 0 match on…

c3ppc3pp:
… delete

### DIFF
--- a/golang/internal/helper/docker/container.go
+++ b/golang/internal/helper/docker/container.go
@@ -15,8 +15,8 @@ import (
 
 const DockerClientTimeoutSeconds = 30
 
-func DeleteContainerByName(ctx context.Context, dog *dogger.DeploymentLogger, nameFilter string) error {
-	matchedContainer, err := GetContainerByName(ctx, nil, nameFilter)
+func DeleteContainerByName(ctx context.Context, dog *dogger.DeploymentLogger, nameFilter string, errorReport bool) error {
+	matchedContainer, err := GetContainerByName(ctx, nil, nameFilter, errorReport)
 	if err != nil {
 		return fmt.Errorf("builder could not get container (%s) to remove: %s", nameFilter, err.Error())
 	}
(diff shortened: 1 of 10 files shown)


golang/internal/helper/docker/container.go
```diff
@@ -54,9 +54,12 @@ func deleteContainerByState(ctx context.Context, dog *dogger.DeploymentLogger, f
 		}
 		return nil
 	case "":
-		return fmt.Errorf(
-			"builder could not determine the state of the container (%s) for deletion, possible theres no container matching the search criteria",
-			filter)
+		if errorReport {
+			return fmt.Errorf(
+				"builder could not determine the state of the container (%s) for deletion, possible theres no container matching the search criteria",
```
nandor-magyar, line 59: theres=>there is

nandor-magyar (approved):
english typos only,  otherwise lgtm

golang/internal/helper/docker/container.go
```diff
@@ -142,41 +145,36 @@ func GetAllContainers(ctx context.Context, dog *dogger.DeploymentLogger) ([]type
...
+	return checkOneContainer(containers, errorReport)
 }
 
 // Making sure there's only one, tops
-func CheckOneContainer(containers []types.Container) (*types.Container, error) {
+func checkOneContainer(containers []types.Container, errorReport bool) (*types.Container, error) {
 	switch len(containers) {
 	case 0:
+		if errorReport {
+			return nil, fmt.Errorf("theres no matching container")
```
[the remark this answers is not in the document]
c3ppc3pp: fixed

golang/internal/helper/docker/container.go
c3ppc3pp: fixed

golang/pkg/dagent/utils/docker.go
```diff
@@ -294,22 +294,19 @@ func DeployImage(ctx context.Context,
 
 	mountList := buildMountList(cfg, dog, deployImageRequest)
 
-	matchedContainer, err := dockerHelper.GetContainerByName(ctx, dog, containerName)
+	matchedContainer, err := dockerHelper.GetContainerByName(ctx, dog, containerName, true)
```
robot9706, line 297: It's not an error if there's no container before deploying the new one. We only need this to report the state of the old container. This is especially important when deploying the very first version of something since the containers won't exist at that point.

I think you could set `errorReport` to false and check `matchedContainer` before writing the state using dogger.

robot9706 (changes requested):
Comment, otherwise LGTM 🚀

golang/pkg/dagent/utils/docker.go
c3ppc3pp: fixed

robot9706 (approved):
LGTM 🚀

PR #362, opened 2022-11-23
fix(crux): increase version fix

peterambrus3:
fix(crux): config mapper and filter for necessary deployments when increasing version
fix(crux-ui): default version cannot be deleted


balanceee (approved):

polaroi8d (approved):
Looks good to me.

balanceee (approved):

PR #367, opened 2022-11-24
fix(crux-ui): nullpointer while focusing an input without editor identity

m8vago:


polaroi8d (approved):
LGTM, tested locally.

balanceee (approved):

PR #368, opened 2022-11-25
refactor(ci): e2e testing

c3ppc3pp:


robot9706 (approved):
LGTM 😎

polaroi8d (approved):
LGTM

web/crux-ui/playwright.config.ts
```diff
@@ -17,8 +17,9 @@ const config: PlaywrightTestConfig = {
     command: 'npm run prod',
     url: baseURL,
     timeout: 60 * 1000, // 1 min
-    reuseExistingServer: !process.env.CI,
+    reuseExistingServer: true,
   },
+  workers: process.env.CI ? 4 : undefined,
```
polaroi8d, line 22: Why we need this?

.github/workflows/full.yml
```diff
@@ -0,0 +1,578 @@
+name: full pipeline
+
+on:
+  push:
+    branches:
+      - develop
+      - main
+  pull_request:
```
polaroi8d, line 8: If `pull_request` is empty why do we need it?

.github/workflows/full.yml
```diff
@@ -0,0 +1,578 @@
...
+      version: ${{ steps.settag.outputs.version }}
+      fallbacktag: ${{ steps.settag.outputs.fallbacktag }}
+    steps:
+      - uses: actions/checkout@v3
+      - uses: dorny/paths-filter@v2
+        id: filter
+        with:
+          filters: |
+            agents:
+              - 'golang/**'
```
polaroi8d, line 62: You can use your predefined env eg.: `GOLANG_WORKING_DIRECTORY` here, no?

.github/workflows/pipeline_set_output_tag.sh
```diff
@@ -21,5 +22,10 @@ if [ $github_ref_type == "tag" ]; then
   VERSION=$github_ref_name
 fi
 
+if [ $github_base_ref == "main" ]; then
```
polaroi8d, line 25: I don't see the source of the `github_base_ref` variable

polaroi8d:
I added a few questions, but overall, good job. Thanks, Matthew, for helping to move forward with the task. 👍🏻

web/crux-ui/playwright.config.ts
c3ppc3pp: So we do not have to write code explicitly for CI, and we can use the same commands as we do during development.

.github/workflows/full.yml
c3ppc3pp: If it is empty, it will match all branches, but without it won't match pull requests.

.github/workflows/pipeline_set_output_tag.sh
c3ppc3pp: The runner sets it, and a bunch of other variables.

.github/workflows/full.yml
c3ppc3pp: Tried, can't. Good catch though, I changed other locations to the variable where I could.

web/crux-ui/playwright.config.ts
polaroi8d: Okey but it would be better if we change `undefined` for some default numbers too.

.github/workflows/full.yml
polaroi8d: Okey, thanks for the clarification.

.github/workflows/full.yml
polaroi8d: Hmmm... Okay, thanks!

.github/workflows/pipeline_set_output_tag.sh
polaroi8d: But you define only the following variables: ```github_ref_type=$1
github_ref_name=$2
github_sha=$3```

github_base_ref undefined in here, no?

web/crux-ui/playwright.config.ts
c3ppc3pp: The `undefined` IS the default. It was shamelessly copied from the official docs: https://playwright.dev/docs/test-parallel#limit-workers

.github/workflows/pipeline_set_output_tag.sh
c3ppc3pp: Well the full case looks like this(also providing some context):
```
      - name: Setting a buildtag
        id: settag
        working-directory: .github/workflows
        run: ./pipeline_set_output_tag.sh $GITHUB_REF_TYPE ${{ github.ref_name }} ${{ github.sha }}
```

The variables here only expanded in various workflow action statements, for example: `run`. It is the first, second and third argument to that shellscript. Since workflow shell can be shaky at times when we do "magic" (additional processing to these variables) we opted for using a separate shellscript so workflow file can be more clean, and the processing part more separated/contained. the $1, $2, $3 are the arguments we pass along, but inside the script we name them again and finally "export" them after processing.

But now we are talking about it... Better add some safeties in the script.

c3ppc3pp: Added safety and yup you were right. I'll sit in the box of shame. image

.github/workflows/full.yml
c3ppc3pp: I was a dumdum, fixed. :))

m8vago:
> I added a few questions, but overall, good job. Thanks @m8vago, for helping to move forward with the task. 👍🏻

Thanks, but it was purely @c3ppc3pp 's catch of the playwright waitForNavigation() misusage, i've just made some lumberjack work. :)

polaroi8d (approved):

web/crux-ui/e2e/utils/common.ts
```diff
@@ -39,7 +40,7 @@ export const extractKratosLinkFromMail = (body: string): string => {
 }
 
 export const kratosFromBaseURL = (baseURL: string) => {
-  const url = replacePort(baseURL, '4434')
+  const url = process.env.KRATOSADMIN_URL || replacePort(baseURL, '4434')
```
m8vago, line 43: that would be KRATOS_ADMIN_URL i think

web/crux-ui/e2e/utils/global-setup.ts
```diff
@@ -39,4 +39,4 @@ const globalSetup = async (config: FullConfig) => {
   await browser.close()
 }
 
-export default globalSetup
+export default globalSetup
```
m8vago, line 42: don't remove empty lines please. They are great friends of @polaroi8d :)

web/crux-ui/e2e/utils/global-teardown.ts
```diff
@@ -33,4 +33,4 @@ const globalTeardown = async (config: FullConfig) => {
   exec('docker rm -f dagent', settings)
 }
 
-export default globalTeardown
+export default globalTeardown
```
m8vago, line 36: empty line missing

web/crux-ui/e2e/verify.spec.ts
```diff
@@ -76,4 +76,4 @@ test.beforeEach(async ({ baseURL, email }) => {
 test.afterEach(async ({ baseURL, email }) => {
   const kratos = kratosFromBaseURL(baseURL)
   deleteUserByEmail(kratos, email)
-})
+})
```
m8vago, line 79: empty line

web/crux-ui/playwright.config.ts
```diff
@@ -8,7 +8,7 @@ const baseURL = process.env.E2E_BASE_URL || 'http://172.17.0.1:8000'
 const config: PlaywrightTestConfig = {
   globalSetup: path.join(__dirname, 'e2e', 'utils', 'global-setup.ts'),
   globalTeardown: path.join(__dirname, 'e2e', 'utils', 'global-teardown.ts'),
-  timeout: 120 * 1000, // 2 min
+  timeout: 2 * 60 * 1000, // 2 min
```
m8vago, line 11: we can adjust this back to 30 sec probably? may 45 if that's more reliable

c3ppc3pp: Yeah, but then we have to drop the worker count to two or one.

web/crux/crux-docker-entrypoint.sh
```diff
@@ -1,5 +1,4 @@
 #!/bin/sh
-export NODE_ENV=production
```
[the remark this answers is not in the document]
c3ppc3pp: Because it will override the variable coming from .env or somewhere else. It will cause issues when you run crux from docker image, and the spawned dagent won't be in debug so it will refuse operations if the GRPC connection is insecure. Basically it sabotages non-prod usage.

web/crux-ui/e2e/utils/common.ts
c3ppc3pp: True, will fix in the pipeline too.

web/crux-ui/e2e/utils/global-setup.ts
c3ppc3pp: Was not intentional. Thanks!

web/crux-ui/playwright.config.ts
m8vago: I think it's fine this way

web/crux-ui/playwright.config.ts
m8vago: whichever is faster overall

web/crux/crux-docker-entrypoint.sh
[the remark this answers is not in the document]
m8vago: is see. we could use it like this:
If it's not defined, we set it to production as a default in the entrypoint otherwise, we use the provided value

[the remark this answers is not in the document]
c3ppc3pp: It is the default, it's defined in the Dockerfile. Thing is we can override it when it's in the Dockerfile, but if we define it inside, then it will override our ENVs.

web/crux-ui/playwright.config.ts
c3ppc3pp: The optimum is somewhere around 2-4 workers, depending on the weather, time, and constellations. Dropped to 2 workers rn, and was timing out on 45s too.

c3ppc3pp: 2 min timeout with 3 workers seems pretty solid. Rebased, let's see it again.

PR #378, opened 2022-12-02
fix(crux): container mapping improvements

m8vago:


web/crux-ui/src/utils.ts
```diff
@@ -310,6 +310,7 @@ export const withContextErrorHandling =
           return redirectTo(ROUTE_INDEX)
         }
 
+        console.error(`[ERROR]: ${err.status} - prop: ${err.property}: ${err.value} - ${err.description}`)
```
peterambrus3, line 313: console.error

m8vago: intentionally

polaroi8d (approved):

PR #386, opened 2022-12-06
fix(agent): add crane deployment list all namespaces

nandor-magyar:
Done:
- create namespace if not exists
- list all deployments/services if namespace is not provided // ports are not yet functional though


polaroi8d (approved):

robot9706 (approved):
LGTM 😎

PR #392, opened 2022-12-07
fix(crux-ui): deployment deploy and log

peterambrus3:
Fix deployment deploy and log functionality.
Also resize the deployment event window.


polaroi8d (approved):
LGTM

PR #393, opened 2022-12-07
build(web): update web deps

m8vago:


c3ppc3pp (changes requested):
I think "chore(web)" would be a more fitting name.

Were the node updates intentional, as you use more recent but not LTS nodeJS versions in the Dockerfiles? Otherwise looks good.

m8vago:
> I think "chore(web)" would be a more fitting name.
> 
> Were the node updates intentional, as you use more recent but not LTS nodeJS versions in the Dockerfiles? Otherwise looks good.

we checked this in conventional commits and it explicitly says build

c3ppc3pp:
> we checked this in conventional commits and it explicitly says build

Alright, then please add this word to `.chglog/config.yml` , so it can show up at the next release!

c3ppc3pp (approved):

web/crux/package.json
```diff
@@ -39,64 +39,64 @@
...
     "ts-node": "^10.9.1",
-    "ts-proto": "1.125.0",
-    "tsconfig-paths": "^4.1.0",
-    "typescript": "^4.7.4"
+    "ts-proto": "^1.135.0",
+    "tsconfig-paths": "^4.1.1",
+    "typescript": "^4.9.3"
   }
-}
+}
```
robot9706, line 102: Missing NL :)

m8vago: :))

polaroi8d (commented):
Ready to review? @m8vago

m8vago:
> Ready to review? @m8vago

nope, still in progress, as the pipeline fails

robot9706 (approved):
LGTM 🚀

PR #399, opened 2022-12-14
feat(web): prevent users from editing registry details

balanceee:
Users can no longer edit registry details (besides name, desc, icon, and the token) if the registry already in use.


web/crux-ui/locales/en/registries.json
```diff
@@ -51,6 +51,9 @@
 
   "altDefaultRegistryPicture": "Picture of registry",
   "errors": {
-    "invalidArgument": "Registry with the given URL doesn't exist"
-  }
+    "invalidArgument": "Registry with the given URL doesn't exist",
```
m8vago, line 54: please try to avoid apostrophes in texts

web/crux/src/app/registry/registry.service.ts
```diff
@@ -87,6 +87,13 @@ export default class RegistryService {
       },
     })
 
-    return this.mapper.detailsToGrpc(registry)
+    const images = await this.prisma.image.count({
```
m8vago, line 90: It could be resolved in the same query
`
    const registry = await this.prisma.registry.findUniqueOrThrow({
      include: {
        _count: {
          select: {
            images: true,
          },
        },
      },
      where: {
        id: req.id,
      },
    })
`

m8vago (approved):

robot9706 (approved):
LGTM 🚀

PR #404, opened 2022-12-15
feat: add LinkAce template

peterambrus3:
https://www.linkace.org/

### DIFF
--- a/web/crux/templates/link-ace.json
+++ b/web/crux/templates/link-ace.json
@@ -1,0 +1,127 @@
+{
+  "name": "LinkAce",
+  "description": "LinkAce is a self-hosted archive to collect links of your favorite websites.\nREQUIREMENTS:\n1) APP_KEY environment variable must be 32 characters long\n2) After deployment, exec in the linkace-app container and add 'chmod 777' privileges to the \\app\\.env file.",
+  "registries": [
+    {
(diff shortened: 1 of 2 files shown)


polaroi8d (approved):

web/crux/templates/link-ace.json
```diff
@@ -0,0 +1,127 @@
...
+        },
+        "name": "linkace-app",
+        "ports": [
+          {
+            "external": 6780,
+            "internal": 80
+          }
+        ],
+        "environment": [
+          {
```
Kovah, line 81: The app key is an important secret used to encrypt/decrypt data in LinkAce. It should probably added as a secret then.

polaroi8d: It's just a template (example), users have to change it when they use their own instance.

polaroi8d: But yea, a good point we'll move this key to the secret.

peterambrus3: Thanks for your review, APP_KEY is a secret from now. 👍

Kovah: Thanks for taking the time to create the template!

balanceee (approved):

PR #417, opened 2022-12-23
[WIP] feat(crux): add backdrop template

peterambrus3:
WIP, backdrop still throws 'Check that the $config_directories variable is correct in settings.php.' exception.


polaroi8d:
I checked the problem and it appears that the backdrop community does not work with Docker and containers. There is no support for the official image. I found a few issues that are 2 years old and still unresolved.

PR #425, opened 2023-01-12
feat(agent): add option not to have CPU limits

nandor-magyar:
Initially we had the incentive to force allocation of every request and limit (CPU and memory). Regarding memory this can be considered a successful practice, but overall machine CPU time could be used more efficiently if services in need could utilize otherwise available CPU capacity.


polaroi8d (approved):
LGTM

robot9706 (approved):
LGTM 🚀

polaroi8d:
I checked the problem and it appears that the backdrop community does not work with Docker and containers. There is no support for the official image. I found a few issues that are 2 years old and still unresolved.

PR #428, opened 2023-01-16
feat(crux): add vaultwarden template

polaroi8d:
Add the Vaultwarden template to the list.


m8vago (approved):

web/crux/templates/vaultwarden.json
```diff
@@ -0,0 +1,42 @@
...
+            "internal": 80
+          }
+        ],
+        "volumes": [
+          {
+            "name": "data",
+            "path": "/data"
+          }
+        ],
+        "environment": [],
```
nandor-magyar, line 33: We need `DOMAIN` environment variable, containing the site URL: https://vault.example.com/.

web/crux/templates/vaultwarden.json
```diff
@@ -0,0 +1,42 @@
...
+  ],
+  "images": [
+    {
+      "name": "vaultwarden-server",
+      "registryName": "Vaultwarden / Docker HUB",
+      "image": "server",
+      "tag": "latest",
+      "config": {
+        "name": "vaultwarden",
+        "ports": [
```
nandor-magyar, line 21: Websocket is also an option for that port 3012 is used, we might not be able to deploy that correctly with crane's current ingress controller features, without modification,
It is possible with dagent though, using custom labels. 
We need a rule like:
```
Host(`vault.dyrectorio.org`) && PathPrefix(`/notifications/hub/negotiate`)
```
That routes to port 3012, all else can go to 80.

nandor-magyar: Custom labels for Traefik to do the custom routing are missing.

polaroi8d: I don't know Traefik labels.

PR #429, opened 2023-01-16
fix(crux-ui): name creation

polaroi8d:
Pull request fixes the name creation error if the user adds whitespace to their name.


robot9706 (approved):
LGTM 📚

PR #431, opened 2023-01-18
fix(ci): conventional commits skip on non-PR runs

c3ppc3pp:


.github/workflows/full.yml
```diff
@@ -35,6 +35,8 @@ jobs:
       - name: Setup node
         uses: actions/setup-node@v2
       - name: Run validation
+        # if it's not a PR we skip
+        if: github.github_base_ref != ''
```
polaroi8d, line 39: Use a more explicit way: `github.event_name == 'pull_request'` or something like that.

c3ppc3pp: Fixed.

polaroi8d (approved):
LGTM

PR #435, opened 2023-01-19
feat(ci): add proto and playwright image builder

c3ppc3pp:


web/crux-ui/e2e/utils/node-helper.ts
```diff
@@ -65,6 +65,13 @@ export const deployWithDagent = async (
   await navigation
 
   if (!ignoreResult) {
+    await page.waitForTimeout(1000)
+    await page.screenshot({ path: screenshotPath('should-successful-after-1k'), fullPage: true })
```
polaroi8d, line 69: Instead of `...-1k` we have to use `...-1s`, so the reader easily can identify the seconds for it, IMHO

polaroi8d (changes requested):

images/proto/package.json
```diff
@@ -0,0 +1,22 @@
+{
```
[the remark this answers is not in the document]
c3ppc3pp: You're damn right. Will fix.

web/crux-ui/e2e/utils/node-helper.ts
c3ppc3pp: These are temporary stuff. Seems we can finally strike off the last flaky test. We're working on it with @m8vago. After I clean up, please remind me to remove these lines if I forget to.

images/proto/package.json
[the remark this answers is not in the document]
c3ppc3pp: Fixed.

web/crux-ui/e2e/utils/node-helper.ts
polaroi8d: Resolved.

polaroi8d (approved):

PR #444, opened 2023-01-27
feat: add optional port binding, alter ui validation

nandor-magyar:
Host binding was inevitable until this change, with this user can expose ports without binding them to host.

The validation was strict on the UI that was lifted, external port can be left empty now, dagent adds these ports to the exposed ports but not the NAT'd ports list. 
Regarding crane this is out of question, portbindings are straight-up ignored.


golang/pkg/crane/k8s/service.go
```diff
@@ -125,7 +125,7 @@ func getServicePorts(portBindings []builder.PortBinding, portRanges []builder.Po
 	ports := []*acorev1.ServicePortApplyConfiguration{}
 
 	for i := range portBindings {
-		portNum := int32(portBindings[i].PortBinding)
+		portNum := int32(*portBindings[i].PortBinding)
```
robot9706, line 128: Does  `PortBinding` always have a value in this case?

m8vago (approved):

polaroi8d (approved):

golang/pkg/crane/k8s/service.go
nandor-magyar: Added fallback: using internal for external if external is not defined.

PR #453, opened 2023-02-03
feat: agent shutdown

robot9706:
- Added `SHUTDOWN` close reason to agent: DAgent tries the DockerAPI first, then a simple process exit
- When a node is deleted the shutdown command is invoked


polaroi8d:
Please add the scope for the PR and fix if it contains only the `dagent` part. Additionally add todo for the crane, please.

golang/pkg/dagent/dagent.go
```diff
@@ -50,6 +51,13 @@ func Serve(cfg *config.Configuration) {
 func grpcClose(ctx context.Context, reason agent.CloseReason) error {
 	if reason == agent.CloseReason_SELF_DESTRUCT {
 		return update.RemoveSelf(ctx)
+	} else if reason == agent.CloseReason_SHUTDOWN {
+		log.Info().Msg("Remote shutdown requested")
+		err := utils.StopSelf(ctx)
+		if err != nil {
+			log.Error().Err(err).Msg("Docker failed to stop")
+			os.Exit(0)
```
nandor-magyar, line 59: Stopping should be done in all environment using Exit 0, is there an issue with it?

robot9706: Docker restarts the container for some reason, but `ContainerStop` keeps it stopped.

nandor-magyar: It is the `unless-stopped` restart policy's behavior, `on-failure` is the one we are looking for if we want 'exit 0' to be a valid termination state for dagent.

robot9706: I see, I'll change the install scripts then, and I'll be able to implement the Kubernetes exit as well I think.

m8vago (approved):

nandor-magyar (approved):
LGTM

PR #454, opened 2023-02-06
feat(crux): reworked email and notification templating

robot9706:
- Removed hardcoded notification and email texts from code and moved them to template files
- Implemented a simple template file
- Moved email and chat templates to a new `assets` folder
- Moved the install scripts from the root folder to the `assets` folder as well
- Fixed an issue with the node script generation form: selecting PowerShell for Docker, then selecting Kubernetes tries to generate a PS1 script for Kubernetes (which is not supported)

❓`assets` folder, template format (inplace + hbr file) /docs for this?/, `notification.template.builder.ts` service (naming?)


web/crux-ui/src/components/nodes/dyo-node-setup.tsx
```diff
@@ -113,6 +113,10 @@ const DyoNodeSetup = (props: DyoNodeSetupProps) => {
             converter={(it: NodeType) => t(`technologies.${it}`)}
             onSelectionChange={it => {
               formik.setFieldValue('type', it, true)
+              if (it === 'k8s') {
```
m8vago, line 116: we should move this if to the onTypeChanged callback, if that's possible

web/crux/src/builders/notification.template.builder.ts
```diff
@@ -0,0 +1,49 @@
...
+
+type TemplateFile = { [key: string]: TemplateEntry }
+
+export type TemplateType = 'email' | 'chat'
+
+export type ProcessedTemplate = { [key: string]: string }
+
+@Injectable()
+export default class NotificationTemplateBuilder {
+  private processTemplateEntry<TData>(entry: TemplateEntry, data: TData): string {
```
m8vago, line 21: simply use T or TemplateType, please

polaroi8d (changes requested):

web/crux/src/builders/notification.template.builder.ts
```diff
@@ -0,0 +1,49 @@
...
+
+    if (!templateText) {
+      throw new InvalidArgumentException({
+        property: 'template',
+        value: 'undefined',
+        message: `Invalid template`,
+      })
+    }
+
+    return Handlebars.compile(templateText)(data)
```
polaroi8d, line 35: "I am not very familiar with Handlebars, but what if it throws an exception? Isn't it necessary to have:

```
    try {
      return Handlebars.compile(templateText)(data);
    } catch (error) {
      throw new Error(`Could not compile template: ${error.message}`);
    }
    ```

File: web/crux/assets/chat/invite.json [hunk 4]
```diff
@@ -0,0 +1,6 @@
+{
```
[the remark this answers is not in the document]
robot9706: Works for me (`notificationTemaple` 😉 ).

m8vago (approved):

polaroi8d (approved):
LGTM

PR #455, opened 2023-02-07
feat(crux): standard health check

robot9706:
- Implemented the standard gRPC health check based on https://github.com/grpc/grpc/blob/master/doc/health-checking.md
- Used `grpc_health_probe` for testing (https://github.com/grpc-ecosystem/grpc-health-probe)


protobuf/proto/health.proto
```diff
@@ -0,0 +1,24 @@
+syntax = "proto3";
```
polaroi8d, line 1: We want to create it independently and not add to the existing services?

robot9706: The standard says that the health check service should be in the 'grpc.health.v1' package, so I think this is the only way I can do that, no?

web/crux/src/app/health/health.controller.ts
```diff
@@ -4,16 +4,65 @@ import { CruxHealthController, CruxHealthControllerMethods, HealthResponse } fro
...
+  constructor(private readonly service: HealthService, private mapper: HealthMapper) {}
 
   @AuditLogLevel('disabled')
   async getHealth(): Promise<HealthResponse> {
     return this.service.getHealth()
   }
+
+  @AuditLogLevel('disabled')
+  async check(request: HealthCheckRequest): Promise<HealthCheckResponse> {
+    if (!request.service || request.service === '') {
```
polaroi8d, line 33: Why do you need to check the `request.service` property falseness?

web/crux/src/app/health/health.controller.ts
```diff
@@ -4,16 +4,65 @@ import { CruxHealthController, CruxHealthControllerMethods, HealthResponse } fro
...
+      const health = await this.service.getHealth()
+      return {
+        status: this.mapper.serviceStatusToProto(health.status),
+      }
+    }
+    if (request.service === 'db') {
```
polaroi8d, line 39: I know this was created earlier but if the `crux` service is down we can't get accurate information from the Database based on the dependencies. At present, it's okay but in the long run, we have to rework this.

web/crux/src/app/health/health.controller.ts
```diff
@@ -4,16 +4,65 @@ import { CruxHealthController, CruxHealthControllerMethods, HealthResponse } fro
...
+      const health = await this.service.getHealth()
+      return {
+        status: health.lastMigration
+          ? HealthCheckResponse_ServingStatus.SERVING
+          : HealthCheckResponse_ServingStatus.NOT_SERVING,
+      }
+    }
+    if (request.service === 'kratos') {
+      const health = await this.service.getKratosHealth()
```
polaroi8d, line 48: You could extract the common variables like `health` outside of the if-else blocks. Minor add new lines between ifs IMHO.

polaroi8d (changes requested):

web/crux/src/app/health/health.controller.ts
robot9706: Rearranged the code a bit.

web/crux/src/app/health/health.controller.ts
```diff
@@ -4,16 +4,62 @@ import { CruxHealthController, CruxHealthControllerMethods, HealthResponse } fro
...
+  constructor(private readonly service: HealthService, private mapper: HealthMapper) {}
 
   @AuditLogLevel('disabled')
   async getHealth(): Promise<HealthResponse> {
     return this.service.getHealth()
   }
+
+  @AuditLogLevel('disabled')
+  async check(request: HealthCheckRequest): Promise<HealthCheckResponse> {
+    if (request.service === '' || request.service === 'db') {
```
m8vago, line 33: this logic should be inside the service itself

web/crux/src/app/health/health.controller.ts
robot9706: I used a standard tool to check the implementation and it sends nothing to check the service health (the standard says empty string so I implemented both).

web/crux/src/app/health/health.controller.ts
```diff
@@ -4,16 +4,50 @@ import { CruxHealthController, CruxHealthControllerMethods, HealthResponse } fro
...
+  @AuditLogLevel('disabled')
+  async check(request: HealthCheckRequest): Promise<HealthCheckResponse> {
+    const status = await this.service.getServiceStatus(request.service)
+    if (status) {
+      return {
+        status: this.mapper.serviceStatusToProto(status),
+      }
+    }
+
+    throw new NotFoundException({
```
m8vago, line 39: move it inside to the serivice

m8vago (changes requested):

m8vago (approved):

PR #460, opened 2023-02-13
fix(crux-ui): volume ids, chips

m8vago:
fixes:
config container and import container volume id’s mixup
config chips are not working
remove capabilities from the ui


web/crux-ui/src/components/products/versions/images/select-images-card.tsx
```diff
@@ -41,7 +41,7 @@ const SelectImagesCard = (props: SelectImagesCardProps) => {
 
   const { data: registries, error: fetchRegistriesError } = useSWR<Registry[]>(API_REGISTRIES, fetcher)
   const [searching, setSearching] = useState(false)
-  const [registry, setRegistry] = useState<Registry>(null)
+  const [registry, setRegistry] = useState<Registry>(registries.length > 0 ? registries[0] : null)
```
robot9706, line 44: This fails because registries is undefined until useSWR loads something.

robot9706 (approved):
Found one bug, otherwise LGTM.

polaroi8d (approved):
LGTM (informal)

PR #464, opened 2023-02-15
fix: fix config mapping, default filters

nandor-magyar:
Testing revealed some issues, here are the fixes forged with @m8vago .

While testing templates found some issues:
- exposing strategy selection was not visible
- default instance view was filtered (ninja fields)

@robot9706 @polaroi8d please check!

### DIFF
--- a/web/crux-ui/src/pages/products/[productId]/versions/[versionId]/deployments/[deploymentId]/instance/[instanceId].tsx
+++ b/web/crux-ui/src/pages/products/[productId]/versions/[versionId]/deployments/[deploymentId]/instance/[instanceId].tsx
@@ -96,6 +96,8 @@ const InstanceDetailsPage = (props: InstanceDetailsPageProps) => {
   }, [editorState.editors])
 
   const onChange = (newConfig: Partial<InstanceContainerConfigData>) => {
+    actions.updateConfig(newConfig)
+
     const newPatch = {
       ...patch.current,
       ...newConfig,
@@ -185,7 +187,7 @@ const InstanceDetailsPage = (props: InstanceDetailsPageProps) => {
 
           {getViewStateButtons()}
         </div>
-        {viewState === 'editor' && <ImageConfigFilters onChange={setFilters} initialBaseFilter="common" />}
+        {viewState === 'editor' && <ImageConfigFilters onChange={setFilters} initialBaseFilter="all" />}
       </DyoCard>
 
       {viewState === 'editor' && (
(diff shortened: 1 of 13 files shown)


web/crux-ui/src/pages/products/[productId]/versions/[versionId]/deployments/[deploymentId]/instance/[instanceId].tsx
```diff
@@ -185,7 +187,7 @@ const InstanceDetailsPage = (props: InstanceDetailsPageProps) => {
 
           {getViewStateButtons()}
         </div>
-        {viewState === 'editor' && <ImageConfigFilters onChange={setFilters} initialBaseFilter="common" />}
+        {viewState === 'editor' && <ImageConfigFilters onChange={setFilters} initialBaseFilter="all" />}
```
Comment (polaroi8d) [inline web/crux-ui/src/pages/products/[productId]/versions/[versionId]/deployments/[deploymentId]/instance/[instanceId].tsx:190]: Why?

web/crux-ui/src/pages/products/[productId]/versions/[versionId]/image/[imageId].tsx
```diff
@@ -194,7 +194,7 @@ const ImageDetailsPage = (props: ImageDetailsPageProps) => {
 
           {getViewStateButtons()}
         </div>
-        {viewState === 'editor' && <ImageConfigFilters onChange={setFilters} />}
+        {viewState === 'editor' && <ImageConfigFilters onChange={setFilters} initialBaseFilter="all" />}
```
Comment (polaroi8d) [inline web/crux-ui/src/pages/products/[productId]/versions/[versionId]/image/[imageId].tsx:197]: Why?

polaroi8d (changes requested):
If @m8vago is a co-author please add his name to the PR description body.

web/crux-ui/src/pages/products/[productId]/versions/[versionId]/deployments/[deploymentId]/instance/[instanceId].tsx
Comment (nandor-magyar) [reply web/crux-ui/src/pages/products/[productId]/versions/[versionId]/deployments/[deploymentId]/instance/[instanceId].tsx:190]: TL;DR When I was testing as a **user** hidden fields in deployments were confusing.

I thought the we have bugs, but the fields that were in use were simply just hidden. On an image screen, it is perfectly fine to have hidden fields, but for a deployment it is obscure, everything that has some values, therefore will be deployed, should be visible. Templates have values that can get lost into the dark.
What do you think?

Comment (polaroi8d) [reply web/crux-ui/src/pages/products/[productId]/versions/[versionId]/deployments/[deploymentId]/instance/[instanceId].tsx:190]: Talked in person, okay for me.

polaroi8d (approved):

PR #467, opened 2023-02-16
feat(web): reset container field buttons

m8vago:


robot9706:
Testing revealed some problems with the config editors:
- When disabling the `Kubernetes` filter, the group still shows but it's empty (I'm guessing it's due to the "extraLBannotations" thing?).
- Args localization is wrong in the filter list.

image

m8vago:
Thanks for the testing, @robot9706 Awesome! I am going to fix them.

PR #469, opened 2023-02-16
fix(crux): various fixes and refinements of templates

nandor-magyar:
Cal.com:
- resource-config
- ReadWrite volume is requested
- DB url corrected
- DB internal expose for verbosity (crane does need it)


robot9706 (approved):
LGTM🐱‍🏍

PR #470, opened 2023-02-16
fix: rename rw to rwo globally

nandor-magyar:
RW is not a valid type name probably modifying crane could have been simpler, but this is a correct choice unfortunately. :crying_cat_face:


nandor-magyar:
As discussed we would rather drop testing databases instead of this custom migration.

robot9706 (approved):
LGTM 🚀

PR #471, opened 2023-02-17
fix: dagent update self container id & ui updating status

robot9706:
- DAgent now uses `HOSTNAME` env (as a container id then name) to try to find its own container (then tries to parse the CGroup file)
- Added a loading indicator to the node update button when node is updating
- Added an `Update in progress` text to the node connection card as well.


web/crux/proto/crux.proto
```diff
@@ -824,6 +825,7 @@ message NodeDetailsResponse {
   optional NodeScriptResponse script = 108;
   optional string version = 109;
   NodeType type = 110;
+    bool updating = 111;
```
nandor-magyar, line 828: Bad indent

nandor-magyar (changes requested):

web/crux/proto/crux.proto
robot9706: oops

polaroi8d (approved):
Tested local, looks okay for me.

nandor-magyar (approved):
LGTM

PR #472, opened 2023-02-17
feat: add technology labels to templates

polaroi8d:
The purpose of this pull request is to address the fact that we currently have templates that are not supported by Kubernetes. We need to find a way to inform users which templates are compatible with which technology.


web/crux-ui/src/components/templates/apply-template-card.tsx
```diff
@@ -35,6 +35,7 @@ const ApplyTemplateCard = (props: ApplyTemplateCardProps) => {
     initialValues: {
       name: propsTemplate.name,
       description: propsTemplate.description,
+      technologies: propsTemplate.technologies,
```
robot9706, line 38: Why the template apply requires the technologies? I thought it's only for UI.

If the `ApplyTemplate` model is the problem, you could do something like:
```
export type ApplyTemplate = Omit<Template, 'technologies'> & {
  type: ProductType
}
```

robot9706 (changes requested):

web/crux-ui/src/components/templates/apply-template-card.tsx
polaroi8d: Yea, that was a problem 🥲 Thanks

robot9706 (approved):
LGTM 🐱‍🏍

PR #482, opened 2023-02-22
feat(cli): add kratos rate limit

c3ppc3pp:


m8vago (approved):

golang/pkg/cli/container_defaults.go
```diff
@@ -311,12 +311,13 @@ func GetKratos(settings *Settings) *containerbuilder.DockerContainerBuilder {
...
-			"traefik.http.routers.kratos.entrypoints":                    "web",
-			"traefik.http.services.kratos.loadbalancer.server.port":      fmt.Sprintf("%d", defaultKratosPublicPort),
-			"traefik.http.middlewares.kratos-strip.stripprefix.prefixes": "/kratos",
-			"traefik.http.routers.kratos.middlewares":                    "kratos-strip",
-			"com.docker.compose.project":                                 cliStackName,
-			"com.docker.compose.service":                                 settings.Containers.Kratos.Name,
+			"traefik.http.routers.kratos.entrypoints":                     "web",
+			"traefik.http.services.kratos.loadbalancer.server.port":       fmt.Sprintf("%d", defaultKratosPublicPort),
+			"traefik.http.middlewares.kratos-strip.stripprefix.prefixes":  "/kratos",
+			"traefik.http.middlewares.kratos-ratelimit.ratelimit.average": "100",
```
nandor-magyar, line 317: This will make the tests fail, not a good idea in general. For rate limits I would just limit the ingress configuration, that is all. If this is absolutely necessary, it has to be configurable and disabled in e2e tests.

c3ppc3pp: We'll see. After the prod-ready compose, I'll try to move the two a bit closer.

PR #484, opened 2023-02-23
feat(crux): add deployment events API & path image

polaroi8d:


web/crux/src/app/deploy/deploy.repository.ts
```diff
@@ -0,0 +1,25 @@
...
+  constructor(private prisma: PrismaService) {}
+
+  async findDeploymentEventsById(deploymentId: string): Promise<any> {
+    const deploymentEvents = await this.prisma.deploymentEvent.findMany({
+      where: {
+        deploymentId,
+      },
+    })
+
+    if (deploymentEvents.length === 0) {
```
m8vago, line 15: I think this is unnecessary. A simple empty array is sufficent.

web/crux/src/app/deploy/deploy.service.ts
```diff
@@ -50,6 +51,7 @@ export default class DeployService {
     private agentService: AgentService,
     imageService: ImageService,
     private mapper: DeployMapper,
+    private repository: DeployRepository,
```
m8vago, line 54: Why do we need this?

web/crux/src/app/image/image.http.controller.ts
```diff
@@ -1,15 +1,19 @@
...
-import { AddImagesToVersionRequestDto, ImageListResponseDto } from 'src/swagger/crux.dto'
+import { AddImagesToVersionRequestDto, ImageListResponseDto, PatchImageRequestDto } from 'src/swagger/crux.dto'
 import JwtAuthGuard from '../token/jwt-auth.guard'
 import ImageService from './image.service'
+import ImagePatchValidationPipe from './pipes/image.patch.pipe'
 
 @Controller('image')
 @UseGuards(JwtAuthGuard)
 @UseInterceptors(HttpLoggerInterceptor)
+@UseFilters(HttpExceptionFilter)
```
m8vago, line 16: Can we put this globally onto the http service itself in the main.ts?

web/crux/src/filters/http-exception.filter.ts
```diff
@@ -0,0 +1,32 @@
...
+export default class HttpExceptionFilter implements ExceptionFilter {
+  catch(exception: Exception, host: ArgumentsHost) {
+    const ctx = host.switchToHttp()
+    const response = ctx.getResponse<Response>()
+    const request = ctx.getRequest<Request>()
+
+    if (exception.name === 'NotFoundException') {
+      return response.status(404).json({
+        statusCode: 404,
+        timestamp: new Date().toISOString(),
```
m8vago, line 19: is this relevant?

m8vago (changes requested):

web/crux/src/filters/http-exception.filter.ts
polaroi8d: Yes, but I can remove.

web/crux/src/app/deploy/deploy.service.ts
polaroi8d: :( service = repository and I would like to improve the layer because mixing the gRPC and HTTP will be backfire.

web/crux/src/app/deploy/deploy.repository.ts
polaroi8d: It's more explicit this way, but I can remove.

web/crux/src/app/image/image.http.controller.ts
polaroi8d: Nope, unused if we are calling the service via gRPC.

web/crux/src/app/deploy/deploy.service.ts
m8vago: :) I can see that we need to improve this, but at the current form this repository just a boilerplate, and does not resolve anything about mixing grpc and http.

Let's discuss this on a call or something.

web/crux/src/app/image/image.http.controller.ts
m8vago: Ah right. I thought nest's http framework is added by some function like the grpc's connectMicroservice() call, but i can see now that is not the case at all, and the app itself is the http framework.

web/crux/src/filters/http-exception.filter.ts
m8vago: No need to remove it, i was just curious.

web/crux/src/app/deploy/deploy.repository.ts
m8vago: I believe we didn't throw any error on other list-returning endpoints for an empty array. Remove it please.

web/crux/src/app/deploy/deploy.service.ts
polaroi8d: Removed.

PR #485, opened 2023-02-23
feat(web): send cookies in grpc metadata

robot9706:
- Removed `accessedBy` field from the Crux gRPC API.
- Crux-UI sends cookies in gRPC Metadata which is then used to get the user ID.
- HTTP API uses user ID from the JWT.


web/crux/src/interceptors/grpc.user.interceptor.ts
```diff
@@ -0,0 +1,49 @@
...
+
+/**
+ * gRPC request user interceptor. Injects the userID into the metadata
+ * based on the cookie from the request.
+ *
+ * Disclaimer: Works only with gRPC controllers.
+ */
+@Injectable()
+export default class GrpcUserInterceptor implements NestInterceptor {
+  private readonly logger = new Logger(GrpcUserInterceptor.name)
```
polaroi8d, line 34: No using of the `logger`.

web/crux/src/services/kratos.service.ts
```diff
@@ -1,15 +1,18 @@
...
-import { Configuration, Identity, IdentityApi, Session } from '@ory/kratos-client'
+import { Configuration, FrontendApi, Identity, IdentityApi, Session } from '@ory/kratos-client'
 import { randomUUID } from 'crypto'
 import { IdentityTraits, KratosInvitation, KRATOS_IDENTITY_SCHEMA } from 'src/shared/models'
 
 @Injectable()
 export default class KratosService {
   private kratos: IdentityApi
 
+  private what: FrontendApi
```
polaroi8d, line 11: `what`?

web/crux/src/interceptors/grpc.user.interceptor.ts
```diff
@@ -0,0 +1,49 @@
...
+import KratosService from 'src/services/kratos.service'
+
+const DISABLE_ACCESSED_BY_METADATA = 'disable-accessed-by-metadata'
+
+export const DisableAccessedByMetadata = () => SetMetadata(DISABLE_ACCESSED_BY_METADATA, true)
+
+export const setAccessedByFromCookie = async (metadata: Metadata, kratos: KratosService) => {
+  const cookie = metadata.getMap().cookie as string
+  if (!cookie) {
+    throw new Error('Cookie missing')
```
polaroi8d, line 14: We can use here the `NotFoundException`, no?

web/crux/src/interceptors/grpc.user.interceptor.ts
```diff
@@ -0,0 +1,49 @@
...
+  }
+
+  const session = await kratos.getSessionByCookie(cookie)
+
+  metadata.add('accessedby', session.identity.id)
```
polaroi8d, line 19: `accessedby` withouth camelCase is okey? Otherhand use const as DISABLED_ACCESSED... would be better, IMHO.

web/crux/src/interceptors/helper.interceptor.ts
```diff
@@ -14,17 +14,19 @@ type GrpcCallLog = {
...
 export default class InterceptorGrpcHelperProvider {
   mapToGrpcObject(context: ExecutionContext): GrpcCallLog {
-    const request = context.getArgByIndex<AccessRequest>(0)
+    const request = context.getArgByIndex(0)
     const surfaceCall = context.getArgByIndex<ServerSurfaceCall>(2)
     return this.mapServerCallToGrpcLog(request, surfaceCall)
   }
 
-  mapServerCallToGrpcLog(request: AccessRequest, surfaceCall: ServerSurfaceCall): GrpcCallLog {
+  mapServerCallToGrpcLog(request: any, surfaceCall: ServerSurfaceCall): GrpcCallLog {
```
polaroi8d, line 22: Can we avoid to using the `any`?

web/crux/src/shared/user-access.guard.ts
```diff
@@ -20,19 +25,22 @@ export default abstract class UserAccessGuard implements CanActivate {
...
-
     if (!request.id) {
       return true
     }
 
-    return this.canActivateWithIdRequest(request)
+    const metadata = context.getArgByIndex<Metadata>(1)
+    const cookie = metadata.getMap().cookie as string
+    if (!cookie) {
+      throw new Error('Cookie missing')
```
polaroi8d, line 35: Throw the deleted Exception or NotFound would be better. 

```throw new InvalidArgumentException({
        message: 'Missing property',
        property: 'accessedBy',
      })```

File: web/crux/src/decorators/grpc.pipe.ts [hunk 7]
```diff
@@ -0,0 +1,27 @@
+import { Metadata } from '@grpc/grpc-js'
```
polaroi8d, line 1: Misleading the decorator folder, its a pipe, no?

web/crux/src/decorators/jwt-user.decorator.ts
```diff
@@ -0,0 +1,8 @@
+import { createParamDecorator, ExecutionContext } from '@nestjs/common'
+
+const JWTUser = createParamDecorator((_: any, context: ExecutionContext) => {
```
polaroi8d, line 3: any? 🌷

polaroi8d (changes requested):

web/crux/src/interceptors/grpc.user.interceptor.ts
m8vago: I think an Unauthorized would be appropiate

web/crux/src/interceptors/helper.interceptor.ts
m8vago: As the AccessRequest was removed, we can not assume any other type than `any`

web/crux/src/shared/user-access.guard.ts
m8vago: Unauthorized please

web/crux/src/interceptors/grpc.user.interceptor.ts
robot9706: For some reason metadata keys are lowercased by default, so even if I used `accessedBy` when trying to get the key it only works with `accessedby`.

Putting it into a const is a good idea though 👍

web/crux-ui/src/components/products/versions/deployments/deployment-view-list.tsx
```diff
@@ -35,7 +35,7 @@ const DeploymentViewList = (props: DeploymentViewListProps) => {
   ]
 
   const itemTemplate = (item: Instance) => [
-    item.image.name,
+    item.image.config.name,
```
m8vago, line 38: we can't be sure it has a config

web/crux/src/decorators/jwt-user.decorator.ts
robot9706: Yes because the request body type is unknown. I could maybe do some generic magic but the request body is not needed anyways.

web/crux/src/app/registry/registry.http.controller.ts
```diff
@@ -20,7 +22,7 @@ export default class RegistryHttpController {
   @ApiBody({ type: AccessRequestDto })
   @ApiOkResponse({ type: RegistryListResponseDto })
   @AuditLogLevel('disabled')
-  async getRegistries(@Body() request: AccessRequest): Promise<RegistryListResponse> {
-    return await this.service.getRegistries(request)
+  async getRegistries(@Body() request: Empty, @JWTUser() accessedBy: string): Promise<RegistryListResponse> {
```
m8vago, line 25: we don't need the empty here

m8vago (approved):

web/crux/src/shared/user-access.guard.ts
```diff
@@ -1,38 +1,41 @@
...
-        message: 'Missing property',
-        property: 'accessedBy',
-      })
+    const metadata = context.getArgByIndex<Metadata>(1)
+    const cookie = metadata.getMap().cookie as string
+    if (!cookie) {
+      throw new UnauthorizedException()
     }
 
-    if (!request.id) {
```
m8vago, line 30: Why did you move this to the concrete access guards?

m8vago (approved):

PR #488, opened 2023-02-28
refactor: container management

m8vago:
Remove containerID from the docker related stack


golang/pkg/dagent/utils/dockerhelper.go
```diff
@@ -7,22 +7,55 @@ import (
...
+			if name == targetName {
+				return &containers[index], nil
+			}
+		}
+	}
+
+	return nil, fmt.Errorf("container not found by prefix(%s), name(%s)", prefix, name)
+}
+
+func DeleteContainerByPrefixName(ctx context.Context, prefix, name string) error {
```
nandor-magyar, line 46: Same

golang/internal/grpc/grpc.go
```diff
@@ -589,44 +593,44 @@ func executeContainerCommand(ctx context.Context, command *common.ContainerComma
 
 func streamContainerLog(reader ContainerLogReader,
 	client agent.Agent_ContainerLogClient,
-	containerID string,
+	prefix, name string,
```
nandor-magyar, line 596: Isn't this the same function that is used for inspecting containers on a node?
Because if that is the case we can't make it sure that a container has a prefix. Either we should not display containers that are not managed by dyrectorio or show logs for other "external" containers too.

golang/pkg/dagent/utils/docker.go
```diff
@@ -305,16 +305,17 @@ func DeployImage(ctx context.Context,
...
 		dog.WriteContainerState("", fmt.Sprintf("Failed to find container: %s", containerName))
 		return err
 	}
+
 	if matchedContainer != nil {
 		dog.WriteContainerState(matchedContainer.State)
 	}
 
-	err = dockerHelper.DeleteContainerByName(ctx, dog, containerName, false)
+	err = dockerHelper.DeleteContainerByID(ctx, dog, matchedContainer.ID)
```
robot9706, line 318: `matchedContainer` can be NIL, which crashes dagent.
I think `DeleteContainerByID` should be moved inside the if above.

web/crux-ui/src/components/products/versions/deployments/deployment-container-status-list.tsx
```diff
@@ -50,51 +53,35 @@ const DeploymentContainerStatusList = (props: DeploymentContainerStatusListProps
       return weak
     }
 
-    const prefix = `${deployment.prefix}-`
-    const containerNames = strong.map(it => {
-      if (it.id && it.name.startsWith(prefix)) {
-        // came from docker
-        return it.name.substring(prefix.length, it.name.length)
-      }
+    const containerNames = strong.map(it => containerNameOf(it.id))
```
robot9706, line 56: I think this is an alread existing bug but I noticed it here:
Containers in the message has no prefix, instead the whole container name comes as `name`, so `containerNameOf` just appends a `-` in front which results in the status list showing the container as not found.

golang/internal/grpc/grpc.go
m8vago: In that case the prefix is a "" string

golang/pkg/dagent/utils/dockerhelper.go
```diff
@@ -7,22 +7,55 @@ import (
...
-	containers, err := dockerHelper.GetAllContainersByName(ctx, nil, nameFilter)
+	containers, err := dockerHelper.GetAllContainersByName(ctx, nameFilter)
 	if err != nil {
 		log.Error().Stack().Err(err).Send()
 	}
 
 	return mapper.MapContainerState(containers)
 }
+
+func GetContainerByPrefixName(ctx context.Context, prefix, name string) (*types.Container, error) {
```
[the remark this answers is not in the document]
nandor-magyar: ?

robot9706 (approved):

PR #491, opened 2023-03-06
feat: add container runtime version checking

c3ppc3pp:
This PR adds container runtime version checking

### DIFF
--- a/golang/internal/runtime/container/container.go
+++ b/golang/internal/runtime/container/container.go
@@ -1,0 +1,135 @@
+package container
+
+import (
+	"context"
+	"errors"
+	"fmt"
+
+	"github.com/docker/docker/api/types"
+	"github.com/hashicorp/go-version"
+	"github.com/rs/zerolog/log"
+)
+
+// If a runtime version doesn't satisfy recommendation: show warning
+// If a runtime version doesn't satisfy minimum: failure
+
+type DockerInfoInterface interface {
+	Info(context.Context) (types.Info, error)
+	ServerVersion(context.Context) (types.Version, error)
+}
+
+const (
+	Podman                         = "podman"
+	Docker                         = "docker"
+	MinimumDockerServerVersion     = "20.10.0"
+	RecommendedDockerServerVersion = "23.0.0"
+	MinimumPodmanServerVersion     = "4.0.0"
+	RecommendedPodmanServerVersion = "4.4.0"
+	PodmanHost                     = "host.containers.internal"
+	DockerHost                     = "host.docker.internal"
+)
+
+var (
+	ErrCannotConnectToServer          = errors.New("cannot connect to server")
+	ErrServerIsOutdated               = errors.New("server is outdated")
+	ErrServerVersionIsNotSupported    = errors.New("serverversion is not supported")
+	ErrServerUnknown                  = errors.New("server is unknown")
+	ErrServerVersionIsNotValid        = errors.New("server version is not valid")
+	ErrCannotParseServerVersion       = errors.New("cannot parse server version")
+	ErrCannotParseVersionConstraint   = errors.New("cannot parse version constraint")
+	ErrCannotGetServerInformation     = errors.New("cannot get server information")
+	ErrCannotGetServerVersion         = errors.New("cannot get server version")
+	ErrCannotSatisfyVersionConstraint = errors.New("cannot satisfy version constraint")
+)
+
+func VersionCheck(ctx context.Context, cli DockerInfoInterface) error {
+	serverVersion, err := cli.ServerVersion(ctx)
+	if err != nil {
+		return err
+	}
+
+	runtime, err := GetContainerRuntime(ctx, cli)
+	if err != nil {
+		return err
+	}
+
+	switch runtime {
+	case Podman:
+		log.Info().Str("version", serverVersion.Version).Msg("Podman")
+
+		err = SatisfyVersion(MinimumPodmanServerVersion, RecommendedPodmanServerVersion, serverVersion.Version)
+		if err != nil {
+			return err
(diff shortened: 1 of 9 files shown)


nandor-magyar (changes requested):
Do not reimplement semver. We already have github.com/hashicorp/go-version added as a dependency.
Check the function `func PreflightChecks` in dagent.

golang/pkg/cli/config_file.go
```diff
@@ -240,22 +252,49 @@ func CheckRequirements(state *Settings) string {
 		log.Fatal().Err(err).Stack().Msg("Cannot get info via docker socket")
 	}
 
-	switch info.InitBinary {
+	return checkVersion(info.InitBinary, info.ServerVersion, state.DisablePodmanChecks)
+}
+
+func checkVersion(init, versionString string, podmanCheck bool) string {
```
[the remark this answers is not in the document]
c3ppc3pp: outdated, thus irrelevant

golang/internal/runtime/container/container.go
```diff
@@ -0,0 +1,120 @@
+package containerruntime
```
nandor-magyar, line 1: container is not enough?

golang/internal/runtime/container/container.go
```diff
@@ -0,0 +1,120 @@
...
+	"context"
+	"fmt"
+
+	"github.com/docker/docker/client"
+	"github.com/hashicorp/go-version"
+	"github.com/rs/zerolog/log"
+)
+
+// If a runtime version doesn't satisfy recommendation: show warning
+// If a runtime version doesn't satisfy minimum: failure\
```
nandor-magyar, line 13: \ at the end?

golang/internal/runtime/container/container.go
```diff
@@ -0,0 +1,120 @@
...
+	Docker                         = "docker"
+	MinimumDockerServerVersion     = "20.10.0"
+	RecommendedDockerServerVersion = "23.0.0"
+	MinimumPodmanServerVersion     = "4.0.0"
+	RecommendedPodmanServerVersion = "4.4.0"
+	PodmanHost                     = "host.containers.internal"
+	DockerHost                     = "host.docker.internal"
+)
+
+func VersionCheck() {
```
nandor-magyar, line 26: ctx and client may be pluggable, for testing and different use cases, where we already have a client or have a valid context already.

golang/internal/runtime/container/container.go
nandor-magyar, line 1: Add some **unit** tests, please.

golang/internal/runtime/container/container.go
nandor-magyar: Also the input should be `serverVersion.Version` for testability.

golang/internal/runtime/container/container.go
```diff
@@ -0,0 +1,120 @@
...
+func GetInternalHostDomain() string {
+	containerRuntime := getContainerRuntime()
+
+	if containerRuntime == Podman {
+		return PodmanHost
+	}
+	return DockerHost
+}
+
+func getContainerRuntime() string {
```
nandor-magyar, line 100: spawning a client/ctx in function makes testing a lot complicated, use parameters.

nandor-magyar (changes requested):

golang/internal/runtime/container/container.go
```diff
@@ -0,0 +1,120 @@
...
+	"github.com/rs/zerolog/log"
+)
+
+// If a runtime version doesn't satisfy recommendation: show warning
+// If a runtime version doesn't satisfy minimum: failure\
+
+const (
+	Podman                         = "podman"
+	Docker                         = "docker"
+	MinimumDockerServerVersion     = "20.10.0"
```
polaroi8d, line 18: We need to perform maintenance on this. Don't we want to use environment variables instead of hardcoded numbers? It would be easier to care about these versions. What do you think? cc. @c3ppc3pp and @nandor-magyar

golang/pkg/cli/config_file.go
```diff
@@ -53,6 +53,12 @@ type Containers struct {
 	MailSlurper    ContainerSettings
 }
 
+type SemVer struct {
```
polaroi8d, line 56: Where re u using these types?

polaroi8d (changes requested):

golang/internal/runtime/container/container.go
nandor-magyar: This is OK as is, the limitations are given and wont change, so these must be constants.

golang/pkg/cli/config_file.go
c3ppc3pp: good catch, removed

golang/internal/runtime/container/container.go
c3ppc3pp: Refactored a bit, thanks!

golang/internal/runtime/container/container.go
c3ppc3pp: Removed.

golang/internal/runtime/container/container.go
c3ppc3pp: Shortened.

golang/internal/runtime/container/container_test.go
```diff
@@ -0,0 +1,52 @@
+package container_test
```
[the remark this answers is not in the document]
nandor-magyar: Still, also make sure your local language server understands build flags in use.

polaroi8d (approved):
(informal) LGTM

nandor-magyar (approved):

PR #497, opened 2023-03-08
feat(crux): maddy template

robot9706:
Added template for the Maddy Email server


robot9706:
Maddy (and email servers in general) requires lots of user interaction to setup, which does not fit with the one-click-deploy idea of dyrector.io. Closing for now.

PR #510, opened 2023-03-21
ci: fix pipeline image push

c3ppc3pp:
Expanded the conditional dependencies at the push steps, fixed a missing env in e2e step.

### DIFF
--- a/.github/workflows/full.yml
+++ b/.github/workflows/full.yml
@@ -5,6 +5,7 @@ on:
     branches:
       - develop
       - main
+    tags:
   pull_request:
 
 permissions:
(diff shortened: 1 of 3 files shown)


.github/workflows/full.yml
```diff
@@ -5,6 +5,7 @@ on:
     branches:
       - develop
       - main
+    tags:
```
polaroi8d, line 8: Is this needed?

c3ppc3pp: Yes. This will trigger the actions on tag pushes. The develop branch builds latest, the main branch builds stable, and tags should build a tagged docker image like "0.4.0" or something.

.github/workflows/full.yml
```diff
@@ -236,12 +215,13 @@ jobs:
 
   crux_test:
     runs-on: ubuntu-22.04
-    needs: [crux_lint, changes]
```
polaroi8d, line 239: Run tests if the lint is failing. Is this a good idea?

.github/workflows/full.yml
polaroi8d: Roger. Resolved.

PR #514, opened 2023-03-23
fix(cli): crux unreachable via localhost using cli

ghost:
When CLI was used crux-ui was throwing internal server error. This was caused by fetching an external url, which doesnt work when you run dyrectorio in containers and connect to it via localhost. This PR fixes this behaviour.

### DIFF
--- a/web/crux-ui/src/utils.ts
+++ b/web/crux-ui/src/utils.ts
@@ -210,7 +210,8 @@ export const configuredFetcher = (init?: RequestInit) => {
 export const fetcher = configuredFetcher()
 
 export const fetchCruxFromRequest = async (req: http.IncomingMessage, url: string, init?: RequestInit) => {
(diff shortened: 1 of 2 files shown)


web/crux-ui/src/utils.ts
```diff
@@ -210,7 +210,8 @@ export const configuredFetcher = (init?: RequestInit) => {
 export const fetcher = configuredFetcher()
 
 export const fetchCruxFromRequest = async (req: http.IncomingMessage, url: string, init?: RequestInit) => {
-  const res = await fetch(`${process.env.CRUX_UI_URL}${url}`, {
+  const externalUrl = process.env.CRUX_URL ? process.env.CRUX_URL : process.env.CRUX_UI_URL
```
m8vago, line 213: process.env.CRUX_URL ?? process.env.CRUX_UI_URL

m8vago (approved):

polaroi8d (approved):

PR #515, opened 2023-03-23
feat(web): storage http api

robot9706:
Implemented HTTP API for storage.


web/crux/src/app/storage/storage.mapper.ts
```diff
@@ -1,28 +1,31 @@
...
+      accessKey: storage.accessKey,
+      secretKey: storage.secretKey,
+      inUse: storage._count?.containerConfigs > 0 || storage._count?.instanceConfigs > 0,
     }
   }
 }
 
 type StorageWithCount = Storage & {
-  _count: {
+  _count?: {
```
m8vago, line 28: why optional?

m8vago (changes requested):

web/crux/src/app/storage/storage.mapper.ts
robot9706: `createStorage` also uses `detailsToDto`, but in that case there's no `_count` since it's new.

m8vago: Create separate type for them please.

m8vago (approved):

PR #516, opened 2023-03-23
fix: dyrectorio compose traefik has missing labels

c3ppc3pp:
Dyrectorio stack's compose file has a traefik, and there was a few new labels introduced, and since there was a new service as well, traefik couldn't deduct which router connects to which service.


nandor-magyar (approved):
LGTM

PR #522, opened 2023-03-29
(refactor): Teams http api

m8vago:


web/crux/src/app/team/team.dto.ts
```diff
@@ -0,0 +1,58 @@
...
+  statistics: TeamStatisticsDto
+}
+
+export class TeamDetailsDto extends TeamDto {
+  users: UserDto[]
+}
+
+export class UpdateUserRoleDto {
+  @ApiProperty({ enum: USER_ROLE_VALUES })
+  role: UserRoleDto
```
robot9706, line 40: `@ValidateNested()` ?

web/crux/src/app/node/node.http.controller.ts
```diff
@@ -26,8 +24,8 @@ export default class NodeHttpController {
    * This is just an experimental implementation, and should be improved in the future.
    */
   @Get('status')
-  @ApiBody({ type: WatchContainerStateRequestDto })
-  @ApiOkResponse({ type: ContainerStateListMessageDto })
+  // @ApiBody({ type: WatchContainerStateRequestDto })
```
robot9706, line 27: This function will be rewritten anyways, so why keep these?

robot9706:
While testing I found that team invites don't work.
When trying to accept an invitation there's an "error:notFound" toast, because `acceptTeamInvitation` is called with PUT instead of POST (I think that's whats happening...).

PR #527, opened 2023-04-03
feat(web): health http api

robot9706:
Moved the health API to HTTP.


web/crux/src/app/health/health.dto.ts
```diff
@@ -0,0 +1,16 @@
...
+
+export class HealthDto {
+  @IsIn(SERVICE_STATUS_VALUES)
+  status: ServiceStatusDto
+
+  @IsString()
+  version: string
+
+  @IsString()
+  @IsOptional()
```
polaroi8d, line 14: Do not need the `@IsOptional()` if we are using `?`.

m8vago: I think it's necessary for the c# generator.

polaroi8d: We don't need.

m8vago (approved):

PR #533, opened 2023-04-12
fix: improve the CLI Docker Engine Server errror message

polaroi8d:


golang/pkg/cli/config_file.go
```diff
@@ -192,10 +192,10 @@ func SettingsFileDefaults(initialState *State, args *ArgsFlags) *State {
 		switch {
 		case errors.Is(err, containerRuntime.ErrServerIsOutdated):
 			NotifyOnce("dockerversion", func() {
-				log.Warn().Stack().Err(err).Msg("Server is outdated, please consider updating")
+				log.Warn().Stack().Err(err).Msg("Docker Engine (Server) is outdated, please consider updating")
```
nandor-magyar, line 195: This branch and the other can also be reached if one is using Podman.
I would opt for Info log level, since it does not affect application behavior:
```
There is a newer version of the container engine in use, please consider updating.
```

golang/pkg/cli/config_file.go
```diff
@@ -192,10 +192,10 @@ func SettingsFileDefaults(initialState *State, args *ArgsFlags) *State {
...
 			})
 		case errors.Is(err, containerRuntime.ErrServerVersionIsNotSupported):
-			log.Fatal().Stack().Err(err).Msg("Server is outdated")
+			log.Fatal().Stack().Err(err).Msg("Docker Engine (Server) is outdated")
```
nandor-magyar, line 198: This also could be Podman as well.

nandor-magyar (changes requested):

nandor-magyar (approved):

PR #550, opened 2023-04-24
fix: respect dynamic email configuration & cli fixes

nandor-magyar:
Done:
- new compose file for plain-text http configurations with warning
- add missing environment variables
- add kratos mail config templating to kratos
- extend .env.example in root


web/docker-compose.dev.yaml
```diff
@@ -1,3 +1,5 @@
+# this file is for development,
+# when crux and crux-ui is executed locally using node
```
m8vago, line 2: NodeJS so it's not confused with dyo's node

m8vago (approved):

m8vago (approved):

PR #556, opened 2023-04-25
feat(ci): add pr labeler

c3ppc3pp:
This PR adds automatic labeling functions to PRs. It will scour the changeset and tries to make an educated guess about what label should be appropriate.

### DIFF
--- a/.github/labeler.yml
+++ b/.github/labeler.yml
@@ -1,0 +1,34 @@
+'source:web': 'web'
+'source:documentation': 'docs'
+'source:agent':
+  - 'golang/pkg/crane'
+  - 'golang/pkg/dagent'
+  - 'golang/cmd/crane'
+  - 'golang/cmd/dagent'
+'source:ci': '.github/workflows'
+'source:build':
+  - '**/Makefile'
+  - 'images'
+  - '**/Dockerfile'
+  - 'golang/build'
+'source:kratos':
+  - 'web/kratos'
+'source:proto':
+  - 'proto'
+  - '**/*.proto'
+'source:cli':
+  - 'golang/pkg/cli'
+  - 'golang/cmd/dyo'
+'lang:makefile': '**/Makefile'
+'lang:sql':
+  - '**/*.prisma'
+  - '**/*.sql'
+'lang:shell':
+  - '**/*.sh'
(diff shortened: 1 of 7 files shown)


.github/labeler.yml
```diff
@@ -0,0 +1,25 @@
...
+'source:documentation': 'docs'
+'source:agent':
+  - 'golang/pkg/crane'
+  - 'golang/pkg/dagent'
+  - 'golang/cmd/crane'
+  - 'golang/cmd/dagent'
+'source:ci': '.github/workflows/'
+'source:build':
+  - '**/Makefile'
+  - 'web/kratos'
```
polaroi8d, line 11: We should add a `source:kratos` for that, IMHO. @m8vago your opinion?

.github/workflows/labeler.yml
```diff
@@ -0,0 +1,12 @@
+name: "Pull Request Labeler"
```
polaroi8d, line 1: We are using the snake case convention to GitHub task names.

.github/labeler.yml
```diff
@@ -0,0 +1,25 @@
+'source:web': 'web'
+'source:documentation': 'docs'
+'source:agent':
+  - 'golang/pkg/crane'
+  - 'golang/pkg/dagent'
+  - 'golang/cmd/crane'
+  - 'golang/cmd/dagent'
+'source:ci': '.github/workflows/'
+'source:build':
```
polaroi8d, line 9: Add `/images` folder for this, please.

polaroi8d (commented):
Small improvements are needed, otherwise, if it works LGTM.

.github/workflows/labeler.yml
c3ppc3pp: Not a single one had snake case :D but aight, changing all

c3ppc3pp: Done!

.github/labeler.yml
c3ppc3pp: Fixed.

.github/labeler.yml
c3ppc3pp: Done!

PR #565, opened 2023-04-27
fix(ci): PR labeling based on changes

c3ppc3pp:
This PR fixes the weird behaviour of the automatic labeler.


polaroi8d (approved):
Okey for me.

PR #576, opened 2023-05-04
fix: nodejs dns resolution

m8vago:
There is a dns resolution error when someone tries to use the cli with `make upd` and uses nodeJS 17+, as the default resolver prefers the ipv6 localhost (`::1`), but our traefik tends to not answer to that. We could investigate further later on @nandor-magyar with a new entrypoint?
However now i just set the dns resolution to prefer ipv4 when it's not a production environment.

- Set dns resolution to prefer ipv4 on non-production nodeJS environments
- Fix some api docs
- Fix GET /node/:nodeId/containers api path


polaroi8d (approved):
LGTM for me.

robot9706 (approved):

PR #577, opened 2023-05-04
fix: production compose database healthcheck user

c3ppc3pp:
This PR fixes database healthchecks in global docker compose files. Also fixes some formatting, and minor things.

### DIFF
--- a/docker-compose.yaml
+++ b/docker-compose.yaml
@@ -2,30 +2,27 @@
 services:
   # If you plan to deploy where a proxy is already in place, feel free to remove the traefik service
   traefik:
+    container_name: traefik
     image: docker.io/library/traefik:2.9
     command:
-      - "--providers.docker=true"
-      - "--providers.docker.exposedbydefault=false"
-      - "--entrypoints.web.address=:80"
-      - "--entrypoints.websecure.address=:443"
-      - "--entrypoints.web.http.redirections.entrypoint.scheme=https"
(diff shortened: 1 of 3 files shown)


docker-compose.yaml
```diff
@@ -173,31 +165,27 @@ services:
...
-
-  # This is insecure, do not use this in production or on the public internet
-  # mailslurper:
-  #   image: docker.io/oryd/mailslurper:smtps-latest
-  #   ports: # port 1025 is used internally for smtp, you do not have to expose that
-  #     - 4436:4436 # web UI
-  #     - 4437:4437 # http API
-  #   restart: unless-stopped
-
+      # This is insecure, do not use this in production or on the public internet
```
m8vago, line 182: this indent is faulty

docker-compose.yaml
```diff
@@ -1,31 +1,28 @@
+---
 # version is removed as per: https://github.com/docker/cli/pull/3445
```
polaroi8d, line 2: Could you remove this comment? I think it's obsoleted, rn.

docker-compose.yaml
polaroi8d: Was this formatted using prettier? I believe the root problem is that you haven't set up prettier properly.

c3ppc3pp: will fix in later PR

docker-compose.yaml
c3ppc3pp: Fixed.

PR #586, opened 2023-05-08
fix(ci): image builder

c3ppc3pp:
This PR resolves some discrepancy around building the "builder image"

This PR resolves the chicken-egg issue in the pipeline by removing the builder image, but using a "bare" VM to install the few stuff we need and build/push/sign the image.

PR #580, opened 2023-05-05
feat: introduce yamlfmt and apply formatting

c3ppc3pp:
This PR adds yaml formatter.

PR #588, opened 2023-05-08
feat(ci): add yaml validation to the pipeline

c3ppc3pp:
As requested: https://github.com/dyrector-io/dyrectorio/pull/580#issuecomment-1536402997

This PR adds YAML validation/linting to the pipeline.

Also fixing broken signing, since cosign had introduced some changes how you should use their software.

PR #580 and #586 are **required** for this.

### DIFF
--- a/.github/workflows/full.yml
+++ b/.github/workflows/full.yml
@@ -22,7 +22,6 @@ env:
   CRUX_UI_WORKING_DIRECTORY: web/crux-ui
   KRATOS_WORKING_DIRECTORY: web/kratos
   GOLANG_WORKING_DIRECTORY: golang
-  BUILDER_IMAGE: ghcr.io/dyrector-io/dyrectorio/alpine-proto:3.17-4
 jobs:
   conventional_commits:
     runs-on: ubuntu-22.04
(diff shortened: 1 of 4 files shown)


.github/workflows/full.yml
```diff
@@ -83,21 +86,19 @@ jobs:
...
           ./check_version.sh ${{ steps.settag.outputs.version }} ../../${{ env.CRUX_WORKING_DIRECTORY }}
           ./check_version.sh ${{ steps.settag.outputs.version }} ../../${{ env.CRUX_UI_WORKING_DIRECTORY }}
-
   # agents scope
   go_lint:
     runs-on: ubuntu-22.04
     needs: gather_changes
     container:
-      image: ghcr.io/dyrector-io/dyrectorio/alpine-proto:3.17-3
+      image: ghcr.io/dyrector-io/dyrectorio/alpine-proto:3.17-4
```
polaroi8d, line 101: Please use a constant instead of these; that way, in the future, we won't need to update all the variables.

.github/workflows/full.yml
```diff
@@ -50,10 +45,18 @@ jobs:
...
-
+  yaml_lint:
+    runs-on: ubuntu-22.04
+    container:
+      image: ghcr.io/dyrector-io/dyrectorio/alpine-proto:3.17-4
+    steps:
+      - name: Checkout
+        uses: actions/checkout@v3
+      - name: Run YAML linting
+        run: yamlfmt -lint .
```
polaroi8d, line 56: minor: add line between jobs please

polaroi8d (changes requested):
Small improvements, otherwise okay for me. Please wait @nandor-magyar's review.

nandor-magyar (approved):
LGTM

nandor-magyar (approved):
LGTM

.github/workflows/full.yml
c3ppc3pp: While defining these can be possible, but rather hacky and work intensive. You cant use expressions there as it turned out.

.github/workflows/full.yml
c3ppc3pp: The tool, yamlfmt, removes empty lines by default, but we can look into this.

polaroi8d: Okey, just we are adding a new line between every job.

polaroi8d: Resolved.

.github/workflows/full.yml
c3ppc3pp: We talked about this IRL, we gonna let it slide for now.

polaroi8d (approved):

PR #594, opened 2023-05-09
feat(crux): add pino logger, and small config refactors

polaroi8d:
PR contains the followings: 
- Implemented Pino logger
- Moved a few files to the config folder
- Small code improvements


web/crux/src/config/pino.logger.config.ts
```diff
@@ -0,0 +1,33 @@
...
+      transport:
+        // Use pino-pretty only in the development environment
+        process.env.NODE_ENV !== 'production'
+          ? {
+              target: 'pino-pretty',
+              options: {
+                singleLine: true,
+                ignore: 'pid,hostname',
+                translateTime: 'mm/dd/yyyy h:MM:ss',
+                messageFormat: '[{req.headers.host}{req.url}][{context}] {msg} [{responseTime} ms][{res.statusCode}]',
```
polaroi8d, line 24: In the long run, we should do this a better way.

web/crux/.env.example
```diff
@@ -13,6 +13,7 @@ JWT_SECRET=jwt-secret-token
 # install script docker pull disabled for the e2e tests
 AGENT_INSTALL_SCRIPT_DISABLE_PULL=true
 
+# Possible values: trace, debug, info, warn, error, and fatal
 LOG_LEVEL=verbose
```
m8vago, line 17: how is verbose the default then?

web/crux/src/config/config.config.ts
```diff
@@ -0,0 +1,6 @@
+const configModuleConfig = {
```
m8vago, line 1: The file name would be nicer just simply as `app.config.ts` IMHO.

web/crux/src/config/pino.logger.config.ts
m8vago: Can we include the `req.method` as well?

m8vago (changes requested):

web/crux/src/config/pino.logger.config.ts
polaroi8d: Not now, we have to figure out the right format for these, but this is not the time 🙉

web/crux/src/config/config.config.ts
polaroi8d: Yeah, good point. I'll rename the object too.

m8vago (approved):

web/crux/.env.example
m8vago: Possible values: trace, debug, info, warn, error, and fatal
there is no verbose in that list

m8vago (changes requested):
Right now it leaks the cookies into the logs. That needs to be resolved before merge.

web/crux/.env.example
polaroi8d: Ahh 🦅

m8vago (approved):

m8vago (approved):

PR #602, opened 2023-05-11
refactor(crux): remove shared module

robot9706:
Remove shared module and move all logic to one level up, also created a new module for audit logging releated services.


m8vago (changes requested):
Discussed some details with the circular deps in person.

m8vago (changes requested):

m8vago (approved):

PR #611, opened 2023-05-15
test(crux): domain agent and deployment tests

robot9706:
Added tests for agent server and deployment mapper.


web/crux/src/domain/agent.spec.ts
```diff
@@ -0,0 +1,380 @@
...
+const AGENT_PUBLIC_KEY = 'public-key'
+
+const GrpcNodeConnectionMock = jest.fn().mockImplementation(() => ({
+  nodeId: AGENT_ID,
+  address: AGENT_ADDRESS,
+  connectedAt: new Date(),
+}))
+
+const observableComplete = <T>(it: Observable<T>) =>
+  new Promise(resolve => {
```
m8vago, line 27: `const subscription = stream.subscribe(() => {})`
`expect(subscription.closed).toBe(true)`

web/crux/src/domain/agent.spec.ts
```diff
@@ -0,0 +1,380 @@
...
+        address: AGENT_ADDRESS,
+        status: 'connected',
+        version: AGENT_VERSION,
+        connectedAt: agentConnection.connectedAt,
+        updating: false,
+      })
+    })
+
+    it('onDisconnected should send unreachable status', async () => {
+      const eventPromise = firstValueFrom(eventChannel.pipe(skip(1)))
```
m8vago, line 85:
```
const resultObs = result.pipe(toArray())
const actual = await firstValueFrom(resultObs)
```

web/crux/src/domain/agent.spec.ts
```diff
@@ -0,0 +1,380 @@
...
+        version: AGENT_VERSION,
+        publicKey: AGENT_PUBLIC_KEY,
+      },
+      eventChannel,
+    )
+  })
+
+  describe('connection status', () => {
+    it('onConnected should send connection status', async () => {
+      const eventPromise = firstValueFrom(eventChannel)
```
m8vago, line 70: you can simply await here

web/crux/src/domain/agent.spec.ts
```diff
@@ -0,0 +1,380 @@
...
+
+      await expect(commandChannel).resolves.toEqual({
+        listSecrets: {
+          prefix: 'prefix',
+          name: 'name',
+        },
+      })
+
+      await expect(secrets).rejects.toThrow()
+    }, 10000)
```
m8vago, line 300: please put a comment here why 10000
export the agent's constant as we discussed in person

web/crux/src/domain/deployment.spec.ts
```diff
@@ -0,0 +1,101 @@
...
+  it('test containerNameFromImageName', () => {
+    expect(containerNameFromImageName('container')).toEqual('container')
+    expect(containerNameFromImageName('/container')).toEqual('container')
+    expect(containerNameFromImageName('/other/container')).toEqual('container')
+
+    expect(containerNameFromImageName('container:tag')).toEqual('container:tag')
+    expect(containerNameFromImageName('/container:tag')).toEqual('container:tag')
+    expect(containerNameFromImageName('/other/container:tag')).toEqual('container:tag')
+
+    // TODO (@robot9706): Real case?
```
m8vago, line 56: nope.

web/crux/src/domain/deployment.spec.ts
```diff
@@ -0,0 +1,101 @@
...
+    expect(containerNameFromImageName('/')).toEqual('/')
+  })
+
+  it('test checkDeploymentCopiability should be true when status is not inProgress or preparing and the version is not rolling', () => {
+    DEPLOYMENT_STATUSES.forEach(deploymentStatus => {
```
m8vago, line 61: check if we can make test cases from an array of values

m8vago (changes requested):

web/crux/src/domain/agent.spec.ts
robot9706: `toArray` requires the subject to complete which in this case it never does. Manually completing it works but it requires more code to do the same thing.

```
it('onConnected should send connection status', async () => {
      const eventPromise = firstValueFrom(eventChannel.pipe(toArray()))

      agent.onConnected()

      eventChannel.complete()
      await expect(eventPromise).resolves.toEqual({
        id: AGENT_ID,
        address: AGENT_ADDRESS,
        status: 'connected',
        version: AGENT_VERSION,
        connectedAt: agentConnection.connectedAt,
        updating: false,
      })
    })
```

m8vago (approved):

PR #616, opened 2023-05-16
chore(crux-ui): remove needless buildstep from dockerfile

c3ppc3pp:
This PR removes an unnecessary buildstep in crux-ui, since we do not use gRPC there.
Also fixed a few image URL-s to fully qualified ones.


web/crux-ui/Dockerfile
```diff
@@ -1,24 +1,16 @@
...
+# stage 1, build app
+FROM docker.io/library/node:19.2-alpine AS BUILDER
 WORKDIR /app
 
 ENV NODE_ENV production
 ENV NEXT_TELEMETRY_DISABLED 1
 
-COPY --from=DEPS /app/node_modules ./node_modules
 COPY . .
+RUN npm ci --no-fund
```
polaroi8d, line 9: We still need to copy `package-lock.json` if we want to use `npm ci` install.

polaroi8d: Nevermind missed the `COPY . .`

polaroi8d (approved):

web/crux-ui/Dockerfile
m8vago: would be nice to copy the package-lock.json in a different layer, so we can reuse those layers while building an image if the dependencies are unchanged.

m8vago (changes requested):

web/crux-ui/Dockerfile
c3ppc3pp: Totally forgot about those! WIll do. Thanks!

c3ppc3pp: Fixed. Also carried this change over to crux. @m8vago please check!

m8vago (approved):

PR #619, opened 2023-05-16
feat(web): add jest to ui and fix the container (internal) missleadings

polaroi8d:
PR contains: 
- Fix the external and internal missleadings
- Add Jest to crux-ui scope
- Add unit test jobs for the pipeline
- Remove unused jobs from the pipeline regarding to crux build
- Add extra dependencies for tests (lint)

### DIFF
--- a/.github/workflows/full.yml
+++ b/.github/workflows/full.yml
@@ -299,7 +299,7 @@ jobs:
         run: npm run lint
   crux_test:
     runs-on: ubuntu-22.04
-    needs: [gather_changes, crux_lint]
+    needs: gather_changes
     defaults:
       run:
         working-directory: ${{ env.CRUX_WORKING_DIRECTORY }}
@@ -362,7 +362,7 @@ jobs:
         run: npm run lint
   crux-ui_unit_test:
     runs-on: ubuntu-22.04
-    needs: [gather_changes, crux-ui_lint]
+    needs: gather_changes
     defaults:
       run:
         working-directory: ${{ env.CRUX_UI_WORKING_DIRECTORY }}
--- a/.github/workflows/full.yml
+++ b/.github/workflows/full.yml
@@ -299,7 +299,7 @@ jobs:
         run: npm run lint
   crux_test:
     runs-on: ubuntu-22.04
-    needs: gather_changes
+    needs: [gather_changes, crux_lint]
     defaults:
       run:
         working-directory: ${{ env.CRUX_WORKING_DIRECTORY }}
@@ -362,7 +362,7 @@ jobs:
         run: npm run lint
   crux-ui_unit_test:
     runs-on: ubuntu-22.04
-    needs: gather_changes
+    needs: [gather_changes, crux-ui_lint]
     defaults:
       run:
         working-directory: ${{ env.CRUX_UI_WORKING_DIRECTORY }}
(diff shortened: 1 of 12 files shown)


web/crux-ui/src/models/container.unit.test.ts
```diff
@@ -0,0 +1,40 @@
+import { ContainerPort, portToString } from './container'
+
+describe('container model tests', () => {
+  beforeEach(() => {})
+
+  it('given both container port when call portToString than add back the full string', () => {
```
robot9706, line 6: `given both container ports portToString should return a full string` maybe?

robot9706 (approved):
Tried running the tests, it worked fine, other than a minor comment it looks good to me.

.github/workflows/full.yml
```diff
@@ -299,7 +299,7 @@ jobs:
         run: npm run lint
   crux_test:
     runs-on: ubuntu-22.04
-    needs: gather_changes
+    needs: [gather_changes, crux_lint]
```
c3ppc3pp, line 302: I'd run the test at the same time as lint/etc runs. Doesn't make much sense logically at first, (because why should we test if the formatting/linting arent good, right?) but it makes the pipeline faster a bit, and checking everything at the same time gives more broad feedback to the dev.

.github/workflows/full.yml
```diff
@@ -364,9 +360,25 @@ jobs:
         run: npm ci --arch=x64 --platform=linuxmusl --cache .npm --prefer-offline --no-fund
       - name: Lint
         run: npm run lint
+  crux-ui_unit_test:
+    runs-on: ubuntu-22.04
+    needs: [gather_changes, crux-ui_lint]
```
c3ppc3pp, line 365: same with the crux_tests

web/crux-ui/src/models/container.unit.test.ts
robot9706: Could work with `calling`

web/crux-ui/package.json
```diff
@@ -25,6 +25,9 @@
     "start": "npm run dev",
     "test:e2e": "playwright test",
     "test:e2e-debug": "playwright test --debug",
+    "test:unit": "jest",
+    "test:unit-watch": "jest --watch",
+    "test:unit-cov": "jest --coverage",
```
m8vago, line 30: Can we match these to crux's package.json commands, or in the other way?

web/crux-ui/src/components/products/versions/version-deployments-section.tsx
```diff
@@ -149,7 +149,6 @@ const VersionDeploymentsSection = (props: VersionDeploymentsSectionProps) => {
             />
           </div>
         )}
-
```
m8vago, line 152: please keep these empty lines, as they are making the code more readable.

web/crux-ui/tsconfig.json
```diff
@@ -30,15 +21,10 @@
     "isolatedModules": true,
     "jsx": "preserve",
     "incremental": true,
-    "typeRoots": ["node_modules/yup"]
+    "typeRoots": ["node_modules/yup"],
+    "outDir": "./dist"
```
m8vago, line 25: we shouldn't really touch tsconfig, cause it's handled by next
i think especially 'outDir' could cause issues.

https://nextjs.org/learn/excel/typescript/create-tsconfig

m8vago (changes requested):

m8vago (approved):

PR #635, opened 2023-05-18
feat(crux): use proxy headers to determine the node ip

m8vago:
Take proxy headers into account while determining the address of a node.

### DIFF
--- a/web/crux/src/main.ts
+++ b/web/crux/src/main.ts
@@ -24,10 +24,8 @@ type GrpcOptions = {
   credentials: ServerCredentials
 }
 
-type GrpcClient = 'api' | 'agent'
-
-const loadGrpcOptions = (certPrefix: GrpcClient, portEnv: string): GrpcOptions => {
-  const port = portEnv ? Number(portEnv) : certPrefix === 'agent' ? 5000 : 5001
+const loadGrpcOptions = (portEnv: string): GrpcOptions => {
+  const port = portEnv ? Number(portEnv) : 5000
 
   return {
     // tls termination occurs at the reverse proxy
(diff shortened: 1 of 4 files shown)


web/crux/src/main.ts
```diff
@@ -58,7 +56,7 @@ const bootstrap = async () => {
   const document = SwaggerModule.createDocument(app, config)
   SwaggerModule.setup('/api/swagger', app, document)
 
-  const agentOptions = loadGrpcOptions('agent', configService.get<string>('GRPC_AGENT_PORT'))
```
polaroi8d, line 61: 🦅

web/crux/src/shared/grpc-node-connection.ts
```diff
@@ -97,6 +101,15 @@ export default class GrpcNodeConnection {
     this.statusChannel.complete()
     this.call.connection = null
   }
+
+  private getFirstItemOfStringArrayMetadata(key: string): string | null {
```
polaroi8d, line 105: Tests? 🤔

web/crux/src/main.ts
m8vago: :man_shrugging:

web/crux/src/shared/grpc-node-connection.ts
m8vago: it's a private method

polaroi8d: https://stackoverflow.com/a/56199483/5804254 ?

polaroi8d (approved):
LGTM

PR #637, opened 2023-05-22
feat(agent): crane pod state mapping

robot9706:
Use the state of pods for the status of a deployment.


golang/internal/mapper/grpc.go
```diff
@@ -514,9 +526,19 @@ func mapServicePorts(svc *corev1.Service) []*common.ContainerStateItemPort {
...
-	if status.ReadyReplicas > 0 {
+func mapKubeStatusToCruxContainerState(state corev1.ContainerState) common.ContainerState {
+	if state.Running != nil {
 		return common.ContainerState_RUNNING
 	}
+	if state.Terminated != nil {
+		return common.ContainerState_EXITED
+	}
+	if state.Waiting != nil {
+		if state.Waiting.Reason == "CrashLoopBackOff" {
```
robot9706, line 537: Which kubernetes statuses should we map? There are lots of them.
Or should we send the state as a string if it's unknown?

golang/internal/mapper/grpc.go
```diff
@@ -434,19 +434,27 @@ func mapContainerPorts(in *[]dockerTypes.Port) []*common.ContainerStateItemPort
...
+	svc *corev1.ServiceList,
+) []*common.ContainerStateItem {
 	stateItems := []*common.ContainerStateItem{}
 	svcMap := createServiceMap(svc)
 
 	for i := range deployments.Items {
 		deployment := deployments.Items[i]
+		pods, podsFound := podsByDeployment[deployment.Name]
+		if len(pods) > 1 {
+			log.Warn().Str("deployment", deployment.Name).Int("numberOfPods", len(pods)).Msg("More than one pod found for deployment")
```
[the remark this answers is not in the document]
polaroi8d: I think, we should implement a custom business logic for that cases. In Kubernetes potentially there will be more than 1 pod in one deployment.

golang/internal/mapper/grpc.go
polaroi8d: I thought we would map the Kubernetes statuses, we should send as specific an error as we can, if we map it to our custom statuses (which are docker related) than it can confuse the users. 

https://github.com/kubernetes/kubernetes/blob/6356023cb42d681b7ad0e6d14d1652247d75b797/pkg/kubelet/container/sync_result.go#L32

golang/pkg/crane/crux/deploy.go
```diff
@@ -31,7 +33,19 @@ func GetDeployments(ctx context.Context, namespace string) []*common.ContainerSt
 		log.Error().Err(err).Stack().Send()
 	}
 
-	return mapper.MapKubeDeploymentListToCruxStateItems(list, svc)
+	podsByDeployment := make(map[string][]corev1.Pod)
```
polaroi8d, line 36: It would be a bad decision to outsource this to a new function? (and also test it with unit tests)

golang/pkg/crane/k8s/deployment.go
```diff
@@ -192,6 +192,22 @@ func (d *Deployment) Scale(namespace, name string, target int) error {
 	return err
 }
 
+func (d *Deployment) GetPods(namespace, name string) ([]coreV1.Pod, error) {
```
polaroi8d, line 195: Tests are missing 🦅

polaroi8d (changes requested):

golang/pkg/crane/k8s/deployment.go
robot9706: I'm not really sure how I'd go about testing this. I added a test yaml with some deployments and the test can check for them, but I think integration tests are not running in the pipeline.

golang/pkg/crane/crux/deploy.go
robot9706: This is the only place where it is currently used, but if you think it'd be a good idea to move it to a separate function I could do it.

golang/Makefile
```diff
@@ -218,6 +218,7 @@ k3d-init-ext: k3d-check
 	k3d cluster create --api-port 172.17.0.1:6443 crane-test && \
 	k3d kubeconfig get crane-test > k3d-auth.yaml && \
 	KUBECONFIG=$(PWD)/k3d-auth.yaml kubectl apply -f https://raw.githubusercontent.com/prometheus-operator/prometheus-operator/main/bundle.yaml
+	KUBECONFIG=$(PWD)/k3d-auth.yaml kubectl apply -f ./crane-test.yaml
```
nandor-magyar, line 221: Revert this please. Add a new make target if you need it. This is used by crane, it is intentional, needed by the stack, if you use prom operator you can spawn service monitors - using its CRDs.

nandor-magyar (changes requested):

golang/Makefile
robot9706: Oh okay I thought this was only used by tests.

nandor-magyar (approved):

polaroi8d (approved):
LGTM (informal)

polaroi8d (approved):
LGTM (informal)

m8vago (approved):

PR #638, opened 2023-05-22
fix: add missing LOG_LEVEL for the project

polaroi8d:


robot9706 (approved):

.env.example
```diff
@@ -22,7 +25,7 @@ KRATOS_SECRET=Random_Generated_String
 # Mailserver settings
 SMTP_URI=smtps://username:password@mailserver.example.com:465
 FROM_EMAIL=from@example.com
-FROM_NAME=dyrectorio Platform
+FROM_NAME="dyrectorio Platform"
```
m8vago, line 28: is it necessary?

docker-compose.yaml
```diff
@@ -25,6 +25,7 @@ services:
     command: [serve]
     environment:
       - TZ=${TIMEZONE:-UTC}
+      - LOG_LEVEL=${LOG_LEVEL:-debug}
```
m8vago, line 28: we have a docker-compose.http-only.yaml as well

m8vago (changes requested):

.env.example
polaroi8d: Auto format. I'll remove it.

docker-compose.yaml
polaroi8d: Oh, yeah, is anyone using that? 🦅

m8vago: yes

m8vago (approved):

PR #639, opened 2023-05-22
fix(dagent): remove agent install non-POSIX parts

nandor-magyar:
We've been setting up docker agents various distros and sites it shows clearly that non-POSIX shell interpreter usage causes problems.

I've ran the shellcheck linter and adjusted the scripts as suggested. In order to avoid future shell issues I would suggest to those who uses other shells on Linux machines than dash/yash to install one of these and override the distribution symlink (in case of Arch especially).  @m8vago @c3ppc3pp

https://marketplace.visualstudio.com/items?itemName=timonwong.shellcheck
https://github.com/itspriddle/vim-shellcheck

### DIFF
--- a/web/crux/assets/install-script/install-docker.sh.hbr
+++ b/web/crux/assets/install-script/install-docker.sh.hbr
@@ -98,7 +98,7 @@ set_environment() {
       HOST_DOCKER_SOCK_PATH="/var/run/docker.sock"
     else
       if [ $(echo "$DOCKER_HOST" | cut -b -7) = "unix://" ]; then
-        HOST_DOCKER_SOCK_PATH="$DOCKER_HOST"
+        HOST_DOCKER_SOCK_PATH=$(echo "$DOCKER_HOST" | cut -b 7-)
       else
         echo "Invalid DOCKER_HOST variable please set HOST_DOCKER_SOCK_PATH if your socket is in a custom location otherwise unset DOCKER_HOST!"
         exit 1


web/crux/assets/install-script/install-docker.sh.hbr
```diff
@@ -93,12 +93,12 @@ set_environment() {
...
-  if [ -z ${HOST_DOCKER_SOCK_PATH:-} ]; then
-    if [ -z ${DOCKER_HOST:-} ]; then
+  if [ -z "${HOST_DOCKER_SOCK_PATH:-}" ]; then
+    if [ -z "${DOCKER_HOST:-}" ]; then
       HOST_DOCKER_SOCK_PATH="/var/run/docker.sock"
     else
-      if [ ${DOCKER_HOST:0:7} = "unix://" ]; then
-        HOST_DOCKER_SOCK_PATH=${DOCKER_HOST:7}
+      if [ $(echo "$DOCKER_HOST" | cut -b -7) = "unix://" ]; then
+        HOST_DOCKER_SOCK_PATH="$DOCKER_HOST"
```
nandor-magyar, line 101: @c3ppc3pp in the original code it copied the initials only with `${DOCKER_HOST:0:7}` was that intentional?

polaroi8d (approved):
LGTM; Tested on 22.3.0 Darwin Kernel Version 22.3.0

robot9706 (approved):
LGTM, tested on Windows 10 with GitBash.

m8vago (approved):

web/crux/assets/install-script/install-docker.sh.hbr
c3ppc3pp: Yeah, HOST_DOCKER_SOCK_PATH is just a path(iirc we use it for traefik deployment, but fix me if im wrong), while DOCKER_HOST is always should contain the protocol too.

web/crux/assets/install-script/install-docker.sh.hbr
```diff
@@ -150,21 +150,21 @@ dagent_install() {
...
     -e TRAEFIK_TLS='true' \
     {{/if}}
-    -e DATA_MOUNT_PATH=$PERSISTENCE_FOLDER \
+    -e DATA_MOUNT_PATH="$PERSISTENCE_FOLDER" \
     {{#if debugMode}}
     -e DEBUG='true' \
     {{/if}}
-    -e HOST_DOCKER_SOCK_PATH=$HOST_DOCKER_SOCK_PATH \
+    -e HOST_DOCKER_SOCK_PATH="$HOST_DOCKER_SOCK_PATH" \
     --add-host=host.docker.internal:host-gateway \
```
c3ppc3pp, line 158: We should wrap this in an if. We shouldn't add this unless its docker && Linux.

web/crux/assets/install-script/install-docker.sh.hbr
```diff
@@ -28,61 +28,61 @@ case $(uname) in
...
     ;;
 esac
 
 set_environment() {
-# Set folder for persisting data
+  # Set folder for persisting data
   PERSISTENCE_FOLDER_FROM_CRUX="{{#if rootPath}}{{{rootPath}}}{{/if}}"
-  if [ -z ${PERSISTENCE_FOLDER:-} ]; then
-    if [ -z ${PERSISTENCE_FOLDER_FROM_CRUX:-} ]; then
+  if [ -z "${PERSISTENCE_FOLDER:-}" ]; then
```
c3ppc3pp, line 47: Error: creating named volume "~/dyrectorio/dagent/": running volume create option: names must match [a-zA-Z0-9][a-zA-Z0-9_.-]*: invalid argument

We should make this into a full path in the script. Or check this in the input field in the webUI

c3ppc3pp (changes requested):

c3ppc3pp (changes requested):
Please clean up around the DOCKER_HOST thing. It is convuluted as heck, I'm open to any solution as long as it's simpler and working correctly.

web/crux/assets/install-script/install-docker.sh.hbr
c3ppc3pp: I'm not sure about this one because this time I defined the directory directly (in the input using web UI), but further down it assigned the ~/.dyrectorio... directory and that was working as it should. I'm fine with it leaving it as it is now, but we should allocate time for this in the near future.

c3ppc3pp:
Tested, apart from the DOCKER_HOST related issue, it works with:
dash-0.5.12
bash-5.2-p15

web/crux/assets/install-script/install-docker.sh.hbr
nandor-magyar: Okay, in order to handle this better, the best we could come up now is adding a hint to prefer absolute paths over relatives and environment variables.

nandor-magyar:
I've used yash 2.54 for testing.

web/crux/assets/install-script/install-docker.sh.hbr
```diff
@@ -150,21 +150,21 @@ dagent_install() {
...
     -e DEBUG='true' \
     {{/if}}
-    -e HOST_DOCKER_SOCK_PATH=$HOST_DOCKER_SOCK_PATH \
+    -e HOST_DOCKER_SOCK_PATH="$HOST_DOCKER_SOCK_PATH" \
     --add-host=host.docker.internal:host-gateway \
     --name 'dagent' \
-    -v $PERSISTENCE_FOLDER/:/srv/dagent \
-    -v $HOST_DOCKER_SOCK_PATH:/var/run/docker.sock \
+    -v "$PERSISTENCE_FOLDER/":/srv/dagent \
+    -v "$HOST_DOCKER_SOCK_PATH":/var/run/docker.sock \
```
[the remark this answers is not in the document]
c3ppc3pp: resolved :heavy_check_mark:

polaroi8d:
Do we need to retest after the new commits?

nandor-magyar:
It was a bit of an edge case, I've tested it, I don't your machines execute the affected branches of the script, so it is fine.
We could just render the script and run it on every potential platform and shell as it is a core feature to start agents.

PR #640, opened 2023-05-23
feat(web): outdated agent

robot9706:
Show "outdated" status for agent. Let users update the agent, and disable every other feature.


web/crux-ui/src/components/nodes/edit-node-card.tsx
```diff
@@ -265,8 +265,9 @@ const EditNodeCard = (props: EditNodeCardProps) => {
                 <DyoButton
                   className="px-6 mt-4 ml-4 mr-auto"
                   secondary
+                  danger={node.status === 'outdated'}
                   onClick={onUpdateNode}
-                  disabled={node.status !== 'connected' || node.updating}
+                  disabled={(node.status !== 'connected' && node.status !== 'outdated') || node.updating}
```
m8vago, line 270: It would be nice to have a pure function in the models folder which contains this logic.

web/crux-ui/src/components/nodes/node-status-indicator.tsx
```diff
@@ -8,6 +8,8 @@ const statusToAssetName = (status: NodeStatus) => {
       return 'circle-red'
     case 'connected':
       return 'circle-green'
+    case 'outdated':
+      return 'circle-turquoise'
```
m8vago, line 12: Turquoise is used as a positive color on buttons. IMHO we should use purple as that is the obsolete deployment's color too.

m8vago (changes requested):

web/crux-ui/src/components/nodes/node-status-indicator.tsx
robot9706: Implemented a `dyo-indicator` component instead of circle svgs, as discussed.

polaroi8d (approved):
LGTM (informal)

web/crux/src/app/agent/agent.service.ts
```diff
@@ -364,8 +359,7 @@ export default class AgentService {
     connection: GrpcNodeConnection,
     request: AgentInfo,
   ): Promise<Observable<AgentCommand>> {
-    const updatedAgent = this.agents.has(request.id)
-    if (updatedAgent) {
+    if (this.agents.has(request.id)) {
       const agent = this.agents.get(request.id)
```
m8vago, line 363: You can just `.get()` the agent and check for` !agent`, so you don't need to lookup the key twice in the hashtable.

m8vago (changes requested):

m8vago (approved):

PR #641, opened 2023-05-24
docs: Use the correct url for app.dyrectorio.com

PoOwAa:
Small fix for README, currently the readme use wrong URL for the SaaS application.


polaroi8d (commented):
Thanks for the fix, @PoOwAa. Please update the commit message (as discussed in person) to align with our policies. Other than that, it looks good to me, thanks for the contribution 🦅

PR #647, opened 2023-05-25
feat: change the version in version.go when using make release command

polaroi8d:


Makefile
```diff
@@ -105,18 +105,20 @@ build-proto-image:
...
 	git pull
 	git checkout -b "release/$(version)"
 
-# ## Create changelog
+## Create changelog
 	git-chglog --next-tag $(version) -o CHANGELOG.md
 	git add CHANGELOG.md
 
+## Change the golang version
+	sed 's/Version *=.*/Version        = "$(version)"/' golang/internal/version/version.go > temp_file && mv temp_file golang/internal/version/version.go
```
polaroi8d, line 120: Need whitespace because of the const format in Golang file.

c3ppc3pp: I'd make changes to the version.go file, so if you add/remove a variable  from that, it won't shift around because of the automatic formatting.

c3ppc3pp (approved):
:+1:

PR #648, opened 2023-05-25
build: add hardcoded agent and cli versions

m8vago:


.github/workflows/check_version.sh
```diff
@@ -1,12 +1,27 @@
...
 
 version=$1
+path=$2
 
-packageversion=$(cat $2/package.json | grep -Eo '\"version\": \"[[:digit:]]+\.[[:digit:]+\.[[:digit:]]+\"')
-formattedversion="\"version\": \"$version\""
-
-if [ "$formattedversion" = "$packageversion" ]; then
-  exit 0
+if [ "${path: -5}" = .json ]; then
```
polaroi8d, line 6: I think you don't need path: -5 easily replaceable with 
```if [[ $path == *.json ]]; then``` same in go IMHO

File: .github/workflows/check_version.sh [hunk 2]
```diff
@@ -1,12 +1,27 @@
...
+elif [ "${path: -3}" = .go ]; then
+  # go
+  package_version=$(cat $path | grep -Eo 'Version[[:blank:]]+= "[[:digit:]]+\.[[:digit:]+\.[[:digit:]]+"')
+  expected_version="Version = \"$version\""
 else
+  echo Invalid file $path
+  exit 2
+fi
+
+#get rid of unnecessary spaces
```
polaroi8d, line 19: minor: comment format, and I don't really get why this is necessary from the comment - maybe its late

.github/workflows/check_version.sh
```diff
@@ -1,12 +1,27 @@
...
+  # json
+  package_version=$(cat $path | grep -Eo '"version": "[[:digit:]]+\.[[:digit:]+\.[[:digit:]]+"')
+  expected_version="\"version\": \"$version\""
+elif [ "${path: -3}" = .go ]; then
+  # go
+  package_version=$(cat $path | grep -Eo 'Version[[:blank:]]+= "[[:digit:]]+\.[[:digit:]+\.[[:digit:]]+"')
+  expected_version="Version = \"$version\""
 else
+  echo Invalid file $path
+  exit 2
```
polaroi8d, line 16: Exit 2 is  `Misuse of shell builtins` its not true here.

.github/workflows/check_version.sh
m8vago: Because in the `version.go` file there are a lot of spaces before the equals sign: `Version             = "0.4.1"`
This line change it to `Version = "0.4.1"`

.github/workflows/check_version.sh
m8vago: Afaik `if [[ ... ]]` is only present in bash, and we like to use just sh.

m8vago: Summon @nandor-magyar

nandor-magyar: You can't do string extrapolation you are using in a POSIX shell. Use cut instead.

nandor-magyar (commented):
Otherwise looking good, missing PR body grr.

.github/workflows/check_version.sh
polaroi8d: Okey. Thanks!

.github/workflows/check_version.sh
polaroi8d: @nandor-magyar which POSIX compatibility checker, did you used before?

PR #651, opened 2023-05-30
fix(crux-ui): deployment list improvement & node connection icon

robot9706:
- Fixed the node icon on the connection card for deployments
- Removed links from the deployment table, removed the eye icon and the whole row navigates to the deployment instead
-


web/crux-ui/src/pages/deployments.tsx
```diff
@@ -77,25 +76,27 @@ const DeploymentsPage = (props: DeploymentsPageProps) => {
...
     clsx('rounded-tl-lg', defaultHeaderClass),
-    ...Array.from({ length: headers.length - 2 }).map(() => defaultHeaderClass),
+    ...Array.from({ length: headers.length - 3 }).map(() => defaultHeaderClass),
     clsx('text-center', defaultHeaderClass),
-    clsx('rounded-tr-lg', defaultHeaderClass),
+    clsx('text-center rounded-tr-lg', defaultHeaderClass),
   ]
 
+  const onCellClick = (data: Deployment, row: number, col: number) => {
+    if (col >= headers.length - 1) {
```
robot9706, line 85: Better method to handle row clicks while handling action column clicks as well?

robot9706: Currently if the user clicks on the row in the actions column that will do nothing, so an action mislick won't navigate.
What do you mean by splitting the second section?

polaroi8d: Okay, never mind - I just misunderstood the question, as I can see now.

web/crux/src/app/deploy/deploy.mapper.ts
```diff
@@ -95,6 +95,7 @@ export default class DeployMapper {
     return {
       id: it.id,
       prefix: it.prefix,
+      note: it.note,
```
polaroi8d, line 98: Could you add unit test for this please 🐒

polaroi8d (approved):

PR #653, opened 2023-05-30
chore(ci): split builder image

c3ppc3pp:
This PR splits up out builder images to be smaller in size, thus improving on pipeline speeds. 

Original builder image was renamed to protobuf, and nodejs, golang, signer images were introduced where name clearly states its intended function. Versioning from now on will be started from 1.


images/builder-protobuf/Dockerfile
```diff
@@ -6,13 +6,15 @@ ENV GOLANGCI_LINT_CACHE $GOPATH/cache
 ENV GOCACHE $GOPATH/cache
 
 SHELL ["/bin/ash","-c"]
+WORKDIR /usr/work
+LABEL org.opencontainers.image.source="https://github.com/dyrector-io/dyrectorio/tree/develop/images/builder-protobuf"
```
polaroi8d, line 10: You are not missing the Dockerfile from the end of the PATH?

images/builder-nodejs/Dockerfile
```diff
@@ -0,0 +1,8 @@
+FROM docker.io/library/node:20-alpine3.17
+
+SHELL ["/bin/ash","-c"]
+WORKDIR /usr/work
+LABEL org.opencontainers.image.source="https://github.com/dyrector-io/dyrectorio/tree/develop/images/builder-nodejs"
+
+RUN apk add alpine-sdk docker git zstd openssl1.1-compat jq
```
polaroi8d, line 7: Sorry for the stupid question but why need git for these images?

.github/workflows/image-builder-golang.yml
```diff
@@ -0,0 +1,64 @@
...
+      - name: artifact upload
+        uses: actions/upload-artifact@v3
+        with:
+          name: builder
+          path: ./builder.zstd
+  push:
+    runs-on: ubuntu-22.04
+    needs: build
+    if: github.ref_name == 'develop'
+    environment: Workflow - Protected
```
polaroi8d, line 37: Why need the environment?

.github/workflows/image-builder-golang.yml
```diff
@@ -0,0 +1,64 @@
...
+    steps:
+      - name: artifact download
```
polaroi8d, line 39: Please be consistent with the names (upper case).

.github/workflows/image-builder-golang.yml
polaroi8d: Ahh, okey - for the secrets from the workflow.

images/builder-protobuf/Dockerfile
```diff
@@ -25,11 +27,6 @@ RUN go install google.golang.org/protobuf/cmd/protoc-gen-go@v1.28 && \
...
 RUN cd /tmp && make install-go-tools && rm Makefile && chmod -R 777 $GOPATH/
 
-RUN curl -s https://raw.githubusercontent.com/k3d-io/k3d/main/install.sh | bash
-RUN curl -L -o /usr/bin/kubectl "https://dl.k8s.io/release/$(curl -L -s https://dl.k8s.io/release/stable.txt)/bin/linux/amd64/kubectl" && chmod +x /usr/bin/kubectl
-
-
-WORKDIR /usr/work
-
-LABEL org.opencontainers.image.source="https://github.com/dyrector-io/dyrectorio/tree/develop/images/alpine-proto"
+#RUN curl -s https://raw.githubusercontent.com/k3d-io/k3d/main/install.sh | bash
```
polaroi8d, line 30: Comments.

.github/workflows/image-builder-golang.yml
```diff
@@ -0,0 +1,64 @@
+name: builder_image_golang_pipeline
```
polaroi8d, line 1: I'm thinking that adding `_pipeline` does not provide any new information to the name. Is this necessary?

images/builder-protobuf/Dockerfile
c3ppc3pp: Not really, as the directory is the whole context of the build. The `docker build` command will accept a directory as parameter and search a Dockerfile in that directory. @nandor-magyar Is this correct this way?

.github/workflows/image-builder-golang.yml
c3ppc3pp: AFAIK no, thx, will remove them.

images/builder-nodejs/Dockerfile
c3ppc3pp: @nandor-magyar using them with golang/Makefile iirc. When I asked about it he stated he want to be able to build a package locally too. For other images seemingly not important.

nandor-magyar: If it is only used for golang? this is the builder-nodejs image.

images/builder-golang/Dockerfile
```diff
@@ -0,0 +1,23 @@
...
+LABEL org.opencontainers.image.source="https://github.com/dyrector-io/dyrectorio/tree/develop/images/alpine-proto"
+
+RUN mkdir -p $GOPATH/src && \
+    mkdir -p $GOPATH/bin && \
+    mkdir -p $GOPATH/cache && \
+    mkdir -p $GOPATH/pkg
+
+RUN apk add alpine-sdk docker git zstd bash
+
+# Need these for integration tests
```
nandor-magyar, line 17: The naming is a bit misleading then, they are also in use as a test runtime, they do more than _build_. Unfortunately if we would want to split container runtime to better isolate concerns (builder/test runtime images), we would be without golang that is used to run tests.

images/builder-protobuf/Dockerfile
nandor-magyar: Yes, this is the correct way, given the spec. This has nothing to do with the docker build command though, this is only for documentation purposes (afaik).

images/builder-golang/Dockerfile
c3ppc3pp: We use these images in our build process(testing is part of the build process), so they got a name like this. If you have something on your mind please tell, I'm open to suggestions.

nandor-magyar: We can live with this, I suppose.

.github/workflows/image-builder-golang.yml
```diff
@@ -0,0 +1,64 @@
...
+  build:
+    runs-on: ubuntu-22.04
+    steps:
+      - name: Checkout
+        uses: actions/checkout@v3
+      - name: Add requirements and build the image
+        run: |
+          sudo apt update
+          sudo apt install zstd golang docker.io containerd runc
+          docker build -f ./images/builder-golang/Dockerfile -t ${GITHUB_REGISTRY}/${BUILDER_IMAGE_NAME}:${VERSION} .
```
nandor-magyar, line 26: The name already speaks two steps (and three is done: setup, build, export). It would be better be split, no?

nandor-magyar (changes requested):

.github/workflows/image-builder-golang.yml
c3ppc3pp: Right. Fixed.

PR #654, opened 2023-05-30
refactor: rename product to project

m8vago:
Rename product to project, as that is less confusing for a new user.


polaroi8d:
Tested local, with my local db. Works fine.

Makefile
```diff
@@ -45,16 +45,6 @@ bundle:
 	zip -r dyrectorio-offline-bundle-$(BUNDLEVER).zip offline
 	mv .env_bak .env || true
 
-## compile docs
```
polaroi8d, line 48: FYI @nandor-magyar, we removed this because we are not updating the docs.

polaroi8d (approved):

PR #657, opened 2023-06-02
feat(web): onboarding

m8vago:
Add the onboarding card to dashboard.


web/crux/src/services/kratos.service.ts
```diff
@@ -117,4 +117,29 @@ export default class KratosService {
     })
     return req.data
   }
+
+  async enableOnboarding(userId: string): Promise<void> {
```
polaroi8d, line 121: What is the motivation to created 2 HTTP API for this?

m8vago: So this way there is no need for a body. it's also more restful IMHO. For example, Spotify uses the same PUT / DELETE on a `:songId/star` resource when you save a song to your liked songs list.

web/crux-ui/src/components/projects/versions/deployments/deployment-container-status-list.tsx
```diff
@@ -72,7 +72,7 @@ const DeploymentContainerStatusList = (props: DeploymentContainerStatusListProps
     const created = new Date(container.date).getTime()
     const seconds = Math.floor((now - created) / 1000)
 
-    const logUrl = deploymentContainerLogUrl(deployment.id, container.id)
+    const logUrl = nodeContainerLogUrl(deployment.id, container.id)
```
robot9706, line 75: This should be `nodeContainerLogUrl(deployment.node.id, container.id)`

robot9706 (changes requested):

robot9706 (approved):

PR #658, opened 2023-06-02
fix: add multi-arch build for CLI and agents

nandor-magyar:
Multi architecture support for agents.


polaroi8d (commented):
Looks okay to me, but we need to test it, I think.

polaroi8d (approved):

PR #660, opened 2023-06-05
fix(crux-ui): add node description to empty node screen

polaroi8d:
Add descriptive text based on user feedback. They don't really know what a node is. I have added a link to provide further information. If it appears too agressive, we can consider removing it.


robot9706:
Fo me it seems a bit off that texts are not centered, but it might be just me.

image

polaroi8d:
> Fo me it seems a bit off that texts are not centered, but it might be just me.
> 
> image

Good point left the `pt-32` there with that auto margin. Fixed 🦅

robot9706 (approved):

m8vago (approved):

PR #663, opened 2023-06-05
build: remove branch from release target

nandor-magyar:
Can be used for hotfixes from now on.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -89,9 +89,18 @@ proto-crux:
 build-proto-image:
 	docker build -t ghcr.io/dyrector-io/dyrectorio/alpine-proto:3.17-4 -f images/alpine-proto/Dockerfile --progress plain .
 
+.PHONY: branch-check
+branch-check:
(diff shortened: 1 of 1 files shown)


Makefile
```diff
@@ -89,12 +89,12 @@ proto-crux:
 build-proto-image:
 	docker build -t ghcr.io/dyrector-io/dyrectorio/alpine-proto:3.17-4 -f images/alpine-proto/Dockerfile --progress plain .
 
+# use on the branch to-release (develop or main for hotfixes)
 .PHONY: release
 release:
-	$(info Do you want to continue? Version will be: $(version))
+	$(info Do you want to continue? Version will be: $(version) from branch: $(shell git rev-parse --abbrev-ref HEAD))
```
polaroi8d, line 95: It would be a nice feature if the shell could display a warning when the branch is not `dev` or `main`, but for now, it's okay with me.

PR #664, opened 2023-06-05
fix(ci): fix the js and tsx source label rules

polaroi8d:
As the title says.


robot9706 (approved):

.github/config/labeler.yml
```diff
@@ -18,4 +20,5 @@ lang:makefile: ['**/Makefile']
 lang:sql: ['**/*.prisma', '**/*.sql']
 lang:shell: ['**/*.sh', '**/*.sh.hbr', '**/*.ps1.hbr']
 lang:golang: ['**/*.go']
-lang:typescript: ['**/*.ts', '**/*.js']
+lang:typescript: ['**/*.ts', '**/*.tsx']
```
nandor-magyar, line 23: `ts?` is not possible?

polaroi8d: what?

nandor-magyar: Use of glob whitecard characters could also be a possibility. 
```
lang:typescript: ['**/*.ts?', '**/*.js?']
```

polaroi8d: If someone adds a random file with a .ts extension, it will be incorrect. I believe explicit types would be better in that case. What are your thoughts @nandor-magyar ?

nandor-magyar: ah, true, tsc, or tsw, or tsa is not a valid extensions, and ? would match those as well.

PR #666, opened 2023-06-06
feat(web): agent audit log

robot9706:
Create an `AgentEvents` table and insert an event when the agent connects, disconnects, kicked (delete or token revoke) or updates.


polaroi8d (changes requested):
When deleting the container, the platform incorrectly states that it was never last connected, which is not true. On the other hand, I believe the logs in the tab where the node isn't connected are invalid because we're not displaying any logs there. We only provide information about the node. In the Figma, we only show agent logs in this tab.

I think we should talk about this in person. 

image

polaroi8d (changes requested):
When deleted a node following error occurs: 

```Foreign key constraint failed on the field: `AgentEvents_nodeId_fkey (index)```

Please try to write E2E test for this. (just add and delete node in the UI)

File: web/crux-ui/src/models/node.ts [hunk 1]
```diff
@@ -66,6 +69,24 @@ export type NodeDeleteContainer = {
   prefix?: string
 }
 
+export type NodeAuditLogQuery = {
+  skip: number
```
m8vago, line 73: This very same schema is used for the auditlogs isn't it? Can we make a PaginationQuery type for this somewhere?

web/crux/prisma/migrations/20230601114252_agentauditlog/migration.sql
```diff
@@ -0,0 +1,2 @@
+-- AlterEnum
+ALTER TYPE "AuditLogContextEnum" ADD VALUE 'agent';
```
m8vago: As discussed in person this is unused.

web/crux/prisma/schema.prisma
```diff
@@ -346,6 +347,7 @@ enum AuditLogContextEnum {
   http
   ws
   rpc
+  agent
```
m8vago: As discussed in person, this is unused.

web/crux/src/app/agent/agent.module.ts
```diff
@@ -31,6 +32,7 @@ import AgentService from './agent.service'
     }),
     ImageModule,
     ContainerModule,
+    AuditLoggerModule,
```
m8vago: unnused

web/crux/src/app/agent/agent.service.spec.ts
```diff
@@ -11,6 +11,7 @@ import { major, minor } from 'semver'
 import AgentService from './agent.service'
 import ContainerMapper from '../container/container.mapper'
 import { NodeConnectionStatus } from '../node/node.dto'
+import AuditLoggerService from '../audit.logger/audit.logger.service'
```
m8vago, line 14: unused

web/crux-ui/src/components/nodes/node-audit-list.tsx
```diff
@@ -0,0 +1,164 @@
...
+  to: Date
+}
+
+interface NodeAuditListProps {
+  node: DyoNode
+}
+
+const defaultHeaderClass = 'uppercase text-bright text-sm font-semibold bg-medium-eased pl-2 py-3 h-11'
+const columnWidths = ['w-2/12', 'w-48', '', 'w-24']
+const sixdays = 1000 * 60 * 60 * 24 * 6 // ms * minutes * hours * day * six
```
m8vago, line 28: sixDays

web/crux/prisma/schema.prisma
```diff
@@ -404,6 +405,16 @@ model NotificationEvent {
   @@unique([event, notificationId])
 }
 
+model AgentEvents {
```
m8vago, line 408: `AgentEvent`

web/crux/src/app/audit/audit.dto.ts
```diff
@@ -33,7 +33,7 @@ export class AuditLogQueryDto extends PaginationQuery {
   readonly to: Date
 }
 
-export const AUDIT_LOG_CONTEXT_VALUES = ['http', 'ws', 'rpc'] as const
+export const AUDIT_LOG_CONTEXT_VALUES = ['http', 'ws', 'rpc', 'agent'] as const
```
m8vago, line 36: `agent` is unused

web/crux/src/app/node/node.service.ts
```diff
@@ -83,6 +86,17 @@ export default class NodeService {
       where: {
         id,
       },
+      include: {
```
m8vago, line 89: Check if this is necessary.

web/crux/src/app/node/node.service.ts
```diff
@@ -314,4 +328,43 @@ export default class NodeService {
...
+          data: true,
+        },
+      }),
+      this.prisma.agentEvents.count({ where }),
+    ])
+
+    return {
+      items: auditLog.map(it => ({
+        ...it,
+        data: it.data as object,
```
m8vago, line 365: Don't we have a type for data?

web/crux/src/interceptors/prisma-error-interceptor.ts
```diff
@@ -92,5 +92,6 @@ export default class PrismaErrorInterceptor implements NestInterceptor {
     NotificationEvent: 'notificationEvent',
     Token: 'token',
     Storage: 'storage',
+    AgentEvents: 'agentEvents',
```
m8vago, line 95: `AgentEvent`

m8vago (changes requested):

web/crux/src/app/node/node.service.ts
robot9706: It's JSON which could contain anything, same as in the AuditLog. I could define a custom type for it, but since we only use it to show it to the user, do we need it?

m8vago (approved):

web/crux/prisma/schema.prisma
```diff
@@ -403,6 +404,16 @@ model NotificationEvent {
...
 }
 
+model AgentEvent {
+  id             String               @id @default(uuid()) @db.Uuid
+  nodeId         String               @db.Uuid
+  createdAt      DateTime             @default(now()) @db.Timestamptz(6)
+  event          AgentEventTypeEnum
+  data           Json?
+
+  node Node @relation(fields: [nodeId], references: [id], onDelete: Cascade)
```
polaroi8d, line 414: Unused space end of the line:`onDelete: Cascade)  `

polaroi8d (approved):

PR #668, opened 2023-06-07
feat(cli): add CLI version in addition to runtime versions

nandor-magyar:
CLI version was missing from the output, minor improvement.

### DIFF
--- a/golang/internal/runtime/container/container.go
+++ b/golang/internal/runtime/container/container.go
@@ -56,7 +56,7 @@ func VersionCheck(ctx context.Context, cli DockerInfoInterface) (*zerolog.Event,
 
 	switch runtime {
 	case Podman:
-		ev := log.Info().Str("Version", serverVersion.Version).Str("Runtime", "Podman")
+		ev := log.Info().Str("Runtime version", serverVersion.Version).Str("Runtime", "Podman")
 
 		err = SatisfyVersion(MinimumPodmanServerVersion, RecommendedPodmanServerVersion, serverVersion.Version)
 		if err != nil {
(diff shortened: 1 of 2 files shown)


.github/workflows/full.yml
```diff
@@ -576,26 +584,54 @@ jobs:
...
+          make GOOS="linux" GOARCHS="amd64 arm64" compile-agents
+      - name: Set up QEMU
+        uses: docker/setup-qemu-action@v2
+      - name: Set up Docker Buildx
+        uses: docker/setup-buildx-action@v2
+      - name: Setup binfmt
+        run: make binfmt
+      - name: Build images
+        run: |
+          ls ./build/out
```
polaroi8d, line 611: ls?

.github/workflows/full.yml
```diff
@@ -608,6 +644,39 @@ jobs:
...
+          path: |
+            $GOPATH/cache
+            $GOPATH/pkg/mod
+          key: ${{ runner.os }}-golang-${{ hashFiles('**/go.sum') }}
+          restore-keys: |
+            ${{ runner.os }}-golang-
+      # fixes: fatal: unsafe repository
+      - name: Adding workspace
+        run: git config --global --add safe.directory "$GITHUB_WORKSPACE"
+      - name: Load go mod
```
polaroi8d, line 661: Mixin `go` ang `Golang`, be consistent please.

golang/Makefile
```diff
@@ -137,27 +138,32 @@ lint:
 .PHONY: build-cli
 build-cli:
 	cd build && \
-	docker build -t ${ORG_REGISTRY}/cli/dyo:$(image_version) -t ${DOCKER_REGISTRY}/dyo:$(image_version) -f cli/Dockerfile .
+	docker buildx build  --build-arg REVISION=$$(git rev-parse --short HEAD) --platform=linux/amd64 --load -t ${ORG_REGISTRY}/cli/dyo:$(image_version) -t ${DOCKER_REGISTRY}/dyo:$(image_version) -f cli/Dockerfile .
+
+.PHONY: build-cli-push
```
polaroi8d, line 143: Don't we want to add a comment for this?

polaroi8d (changes requested):

.github/workflows/full.yml
nandor-magyar: Removed the change.

.github/workflows/full.yml
nandor-magyar: Golang is the language, go.mod is the lock file, but could be more verbose.

m8vago: I think it's fine

polaroi8d (approved):

PR #673, opened 2023-06-08
fix: registry and storage editing

m8vago:
Remove the editing restrictions from registries and storages, when they are in use.


nandor-magyar (approved):
LGTM! Nice cleanup. :+1:

PR #676, opened 2023-06-09
fix(web): deployment environment editing

m8vago:
- Fixes an issue, where the empty line wasn't added to the deployment environment list while editing it.
- Fixes instance config and image config resetSection patch distribution.


polaroi8d (approved):
LGTM

PR #677, opened 2023-06-09
fix(dagent): add import container logs parsing & send

nandor-magyar:
Import container output was missing, regardless of the exit code now it is sent with the status.

Moved the `ReadDockerLogsFromReadCloser` function next-to other reusable functions.
@robot9706 it was nice that we have this already. :+1: 
Note that I've extended it a bit, please check it out.


polaroi8d (approved):
LGTM

PR #678, opened 2023-06-12
fix(web): rolling version deployments

robot9706:
Rolling versions can only have one deployment for a node with the same prefix.


web/crux/src/app/deploy/interceptors/deploy.create.interceptor.ts
```diff
@@ -26,7 +38,7 @@ export default class DeployCreateValidationInterceptor implements NestIntercepto
 
     if (existingDeployment) {
       throw new CruxConflictException({
-        message: 'There is already a deployment with preparing status for the version on that node',
+        message: version.type === VersionTypeEnum.rolling ? 'rollingVersionDeployment' : 'alreadyHavePreparing',
```
m8vago, line 41: Leave the message as an information for the developer.
Use a property and value instead. `property: 'version.type', value: `rolling`

m8vago (changes requested):

web/crux/src/app/deploy/interceptors/deploy.create.interceptor.ts
robot9706: Unfortunatelly value must be the existingDeployment id because we need to redirect the user to the already existing deployment.

m8vago: As we discussed in person, you could add an `error?: string` field for this purposes.

m8vago (changes requested):

m8vago (approved):

PR #682, opened 2023-06-13
build: remove signing from multiarch images

nandor-magyar:
Multiarch images seem to be overwritten by non-multiarch images, this might be the culprit.


polaroi8d (approved):
Please add the motivation for why we commented out.

m8vago (approved):

PR #686, opened 2023-06-13
feat(crux-ui): add tooltips and refactor <Image> to <DyoIcon>

polaroi8d:
As the title says, I refactored the affected <Image> tags to <DyoIcon> and started using the `alt` attribute for the tooltips. However, there are still a few leftovers that are using the <Image> tag.


nandor-magyar (approved):
LGTM

PR #695, opened 2023-06-16
build(ci): add missing go_push requirement

nandor-magyar:
Golang images push job is now skipped because it is executed too early.


polaroi8d (approved):
🦅

PR #697, opened 2023-06-16
feat(crane): add explicit ClusterRoleBinding and explicit Rules

nandor-magyar:
Originally we hardwired cluster-wise admin access, with this change the install script attempts to require the least amount of access necessary to manage resources.
Resources that are read/create only:
- namespaces
- PVCs


robot9706 (approved):
🏗

PR #706, opened 2023-06-21
test(crux-ui): add e2e test for version and deployment deletes and fix test flakyness

m8vago:
- Added default version tests
- Added dagent log to e2e artifacts (+ kratos stderr)


web/crux-ui/e2e/with-login/resource-copy.spec.ts
```diff
@@ -0,0 +1,344 @@
...
+    await settingsButton.click()
+
+    const sock = waitSocket(page)
+    await page.waitForSelector(`h2:has-text("Container")`)
+    const ws = await sock
+    const wsRoute = deploymentWsUrl(defaultDeploymentId)
+
+    const internal = '1000'
+    const external = '2000'
+    await addPortsToContainerConfig(page, ws, wsRoute, internal, external)
```
robot9706, line 190: This fails because `addPortsToContainerConfig` uses `wsPatchSent` which waits for `WS_TYPE_PATCH_IMAGE` but in this case it's an instance patch so it never receives the expected frame :/

polaroi8d (approved):
LGTM (Informal)

PR #716, opened 2023-06-26
feat(web): token never expiration

robot9706:
Added 'Never' option for user API tokens.


web/crux-ui/src/components/settings/create-token-card.tsx
```diff
@@ -6,15 +6,15 @@ import { DyoHeading } from '@app/elements/dyo-heading'
...
+import { GeneratedToken, GenerateToken, TokenExpiration } from '@app/models'
 import { API_TOKENS } from '@app/routes'
 import { sendForm } from '@app/utils'
 import { generateTokenSchema } from '@app/validations/token'
 import { useFormik } from 'formik'
 import useTranslation from 'next-translate/useTranslation'
 import { MutableRefObject } from 'react'
 
-const EXPIRATION_VALUES = [30, 60, 90]
+const EXPIRATION_VALUES: TokenExpiration[] = [30, 60, 90, 'never']
```
polaroi8d, line 17: Not better to use 0 for never, and we can skip check like these `typeof value === 'number'` ?

robot9706: I wanted to be explicit about it, but 0 could work too.

polaroi8d: Mixing strings with numbers not a good approach IMHO.

web/crux/prisma/schema.prisma
```diff
@@ -28,12 +28,12 @@ model Team {
...
-  userId    String   @db.Uuid
-  name      String   @db.VarChar(70)
-  createdAt DateTime @default(now()) @db.Timestamptz(6)
-  expiresAt DateTime @db.Timestamptz(6)
-  nonce     String   @db.Uuid
+  id        String    @id @default(uuid()) @db.Uuid
+  userId    String    @db.Uuid
+  name      String    @db.VarChar(70)
+  createdAt DateTime  @default(now()) @db.Timestamptz(6)
+  expiresAt DateTime? @db.Timestamptz(6)
```
polaroi8d, line 35: Nullable?

polaroi8d (approved):
One minor remark, but LGTM.

PR #718, opened 2023-06-27
feat: use events to track container states in agents

robot9706:
Instead of polling containers use docker events to only track changes and notify crux if something happened.


golang/internal/mapper/grpc.go
```diff
@@ -375,46 +375,48 @@ func mapVolumeLinks(in []*agent.VolumeLink) []v1.VolumeLink {
...
+		ImageTag:  imageTag,
+	}
+}
 
-		if len(imageName) > 1 {
-			imageTag = imageName[1]
-		} else {
-			imageTag = "latest"
-		}
+func MapContainerStates(in []dockerTypes.Container, prefix string) []*common.ContainerStateItem {
```
m8vago, line 414: `MapContainerStateList` so we're not going to confuse it with `MapContainerState`

golang/pkg/crane/crux/deploy.go
```diff
@@ -16,36 +17,56 @@ import (
...
-		pods, err := deploymentHandler.GetPods(namespace, deployment.Name)
-		if err != nil {
-			log.Error().Err(err).Stack().Send()
-			return nil
-		}
+	eventChannel := make(chan []*common.ContainerStateItem)
 
-		podsByDeployment[deployment.Name] = pods
-	}
+	// TODO: Proper kubernetes event handling
```
m8vago, line 29: Let's check if we can use kube's event api with the current .proto changes in the future.

golang/pkg/dagent/utils/dockerhelper.go
```diff
@@ -5,34 +5,136 @@ package utils
...
 
+	containerID := common.ContainerIdentifier{
+		Prefix: prefix,
+		Name:   name,
+	}
+
+	return &containerID
+}
+
+func eventToMessage(ctx context.Context, prefix string, event *events.Message) (*common.ContainerStateItem, error) {
```
m8vago, line 46: Please rename it to something, which contain the actual struct names.

golang/pkg/dagent/utils/dockerhelper.go
```diff
@@ -5,34 +5,136 @@ package utils
...
+			Ports:     []*common.ContainerStateItemPort{},
+			ImageName: "",
+			ImageTag:  "",
+		}
+		return destroy, nil
+	}
+
+	containerState := mapper.MapDockerContainerEventToContainerState(event.Action)
+	if containerState == common.ContainerState_CONTAINER_STATE_UNSPECIFIED {
+		return nil, nil
```
m8vago, line 77: Let's write a comment about why we ignore some events, and provide some examples.

golang/pkg/dagent/utils/dockerhelper.go
```diff
@@ -5,34 +5,136 @@ package utils
...
+		return nil, err
+	}
+
+	newState := mapper.MapContainerState(container, prefix)
+	newState.State = containerState
+	return newState, nil
+}
+
+func WatchContainersByPrefix(ctx context.Context, prefix string) (*grpc.ContainerWatchContext, error) {
+	cli, err := client.NewClientWithOpts(client.FromEnv, client.WithAPIVersionNegotiation())
```
nandor-magyar, line 91: This is better as a parameter. You might need to split this into two because of this.

web/crux/src/domain/container-status-watcher.ts
```diff
@@ -1,18 +1,23 @@
...
 import { CruxPreconditionFailedException } from 'src/exception/crux-exception'
 import { AgentCommand } from 'src/grpc/protobuf/proto/agent'
-import { ContainerStateListMessage } from 'src/grpc/protobuf/proto/common'
+import { ContainerState, ContainerStateItem, ContainerStateListMessage } from 'src/grpc/protobuf/proto/common'
 import GrpcNodeConnection from 'src/shared/grpc-node-connection'
+import { Agent } from 'src/domain/agent'
 
 export type ContainerStatusStreamCompleter = Subject<unknown>
 
+type StateMap = { [containerIdentifier: string]: ContainerStateItem }
```
m8vago, line 10: `Record<string, ContainerStateItem>`

golang/pkg/dagent/utils/dockerhelper.go
```diff
@@ -5,34 +5,136 @@ package utils
...
+	if containerID == nil {
+		return nil, errors.New("event has no container name")
+	}
+
+	if prefix != "" && containerID.Prefix != prefix {
+		return nil, nil
+	}
+
+	if event.Action == "destroy" {
+		destroy := &common.ContainerStateItem{
```
nandor-magyar, line 61: Since you are returning this right away, you could inline it into the return.

nandor-magyar (changes requested):

web/crux/src/domain/container-status-watcher.ts
```diff
@@ -30,7 +35,29 @@ export default class ContainerStatusWatcher {
...
+        (map, it) => ({
+          ...map,
+          [it]: this.state[it],
+        }),
+        {},
+      )
+    this.state = updated.reduce(
+      (map, it) => ({
+        ...map,
+        [Agent.containerPrefixNameOf(it.id)]: it,
```
m8vago, line 55: Let's try to evade of creating new objects.

m8vago (changes requested):

m8vago (approved):

nandor-magyar (approved):

PR #721, opened 2023-07-03
feat(dagent): add useParent functionality to dagent

nandor-magyar:
It was implemented a while ago in our crane, but it was missing from dagent.


polaroi8d (approved):
LGTM (informal)

golang/internal/helper/image/image.go
```diff
@@ -204,49 +205,80 @@ func Pull(ctx context.Context, cli client.APIClient, logger io.StringWriter, exp
...
-	cli, cliErr := client.NewClientWithOpts(client.FromEnv, client.WithAPIVersionNegotiation())
-	if cliErr != nil {
-		return cliErr
-	}
-
-	if !forcePull {
+func shouldUseLocalImage(ctx context.Context, cli client.APIClient,
+	distributionRef reference.Named, encodedAuth string, preferLocal bool,
+) (bool, error) {
+	if preferLocal {
```
robot9706, line 225: Feels a bit strange calling this function only to check a param, while you could check it before calling it.

Also `preferLocal` feels like it should only be considered if both local and remote images exist and then local should be used. So maybe `forceLocal` or something like that?

golang/pkg/dagent/utils/init_container.go
```diff
@@ -45,16 +51,14 @@ func spawnInitContainer(
 
 	resultCont, waitResult, err := builder.
 		WithClient(cli).
-		WithImage(cont.Image).
-		WithCmd(cont.Command).
+		WithImage(initCont.Config.Name).
+		WithCmd(initCont.Config.Command).
 		WithName(initContName).
-		WithEnv(EnvMapToSlice(cont.Envs)).
+		WithEnv(initCont.EnvList).
```
robot9706, line 57: We should use both InitContainer envs + parent config envs.

robot9706 (changes requested):

golang/pkg/dagent/utils/init_container.go
nandor-magyar: Nice catch.

golang/internal/helper/image/image.go
nandor-magyar: As discussed it is prefer local because it attempts to use remote if local is not available.

robot9706:
Tested, init containers seem to use their name as their image name, and the containers do not actually have any environment variables.

robot9706 (approved):
Tested, seems to be working fine 🎈

PR #727, opened 2023-07-05
fix(dagent): if useParent is toggled containers share network too

nandor-magyar:
Internal ports are not available if networks are not shared with init containers.


robot9706 (approved):

golang/pkg/builder/container/container_builder.go
```diff
@@ -64,7 +66,7 @@ type DockerContainerBuilder struct {
 	ctx             context.Context
 	client          client.APIClient
 	containerID     *string
-	networkIDs      []string
+	networkMap      map[string]string
```
polaroi8d, line 69: Does this change affect any other module?

nandor-magyar: I am not aware, this was internal, no usage outside of the package, the only use of the change is to better inform the user about what is happening.

PR #728, opened 2023-07-05
refactor(crux): use @updatedAt without optional in prisma schema

AmorfEvo:
All `updatedAt` fields are annotated with `@updatedAt`, so the unnecessary optional marks are removed.
This annotation automatically stores the time when a record was last updated.


nandor-magyar (commented):
LGTM. @robot9706 have a look as well :cat2:

nandor-magyar (approved):

polaroi8d (approved):

PR #733, opened 2023-07-06
refactor(crux): refactor http response status codes for request handlers

AmorfEvo:
We use the built-in Nest.js HttpStatus enums as @HttpCode parameters instead of magic numbers.


robot9706 (approved):

polaroi8d (approved):

polaroi8d (changes requested):
Missing the scope and the type of the PR in the commit message.

polaroi8d (approved):
LGTM now, thanks.

PR #734, opened 2023-07-07
cicd: add deploy_external workflow for external webhooks & sundyo

nandor-magyar:
This is an attempt of dogfooding, where we use github actions to trigger our CD for rolling version, where we update a remotely deployed dyrector.io :pizza:

### DIFF
--- a/.github/workflows/deploy_external.yaml
+++ b/.github/workflows/deploy_external.yaml
@@ -1,0 +1,23 @@
+# workflow containing webhooks that are external to us
+name: deploy_external
+on:
+  workflow_run:
+    workflows:
+      - "product_builder"
+    branches:
+      - "develop"
+    types:
+      - completed
+permissions:
+  contents: read
+  pull-requests: read
+  actions: read
+jobs:
+  execute-webhook:
+    if: ${{ github.event.workflow_run.conclusion == 'success' }}
+    runs-on: ubuntu-22.04
+    steps:
+      - name: Webhook
+        run: |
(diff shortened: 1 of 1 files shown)


.github/workflows/deploy_external.yaml
```diff
@@ -0,0 +1,23 @@
...
+  pull-requests: read
+  actions: read
+jobs:
+  execute-webhook:
+    if: ${{ github.event.workflow_run.conclusion == 'success' }}
+    runs-on: ubuntu-22.04
+    steps:
+      - name: Webhook
+        run: |
+          curl -X POST -H 'Authorization: Bearer ${{SUN_DYO_WEBHOOK_TOKEN}}' https://app.dyrectorio.com/api/deployments/8322f763-e611-40c7-a29b-bb21f2fd4822/start
```
polaroi8d, line 22: Don't we want to add this to GitHub `env` as well? If the deployment changes in the future, we can easily modify the variable.

.github/workflows/deploy_external.yaml
```diff
@@ -0,0 +1,23 @@
...
+permissions:
+  contents: read
+  pull-requests: read
+  actions: read
+jobs:
+  execute-webhook:
+    if: ${{ github.event.workflow_run.conclusion == 'success' }}
+    runs-on: ubuntu-22.04
+    steps:
+      - name: Webhook
```
polaroi8d, line 20: What if we have 3 or more external instances? What will be their names? I think we can include the name of the company or use some type of code to avoid exposing any sensitive data.

nandor-magyar: Modified as discussed.

.github/workflows/deploy_external.yaml
nandor-magyar: It will stay as is, as discussed.

polaroi8d (approved):
LGTM, cool!

PR #740, opened 2023-07-11
fix(crux): deployment token name validation

m8vago:
Fix an issue where you can't make a deployment token, because the name validation was wrong.


robot9706 (approved):
Tested, works.

polaroi8d (approved):

PR #743, opened 2023-07-11
fix(crux-ui): add view buttons to deployment list rows

AmorfEvo:
- Fixed view buttons were missing from the deployments list rows.
- Fixed project list rows were not clickable to view them (similarly to the deployment list rows).
- Fixed version deployment list rows were not clickable to view them (similarly to the deployment list rows).


robot9706:
Tested, cursor is not correct when hovering a deployment row in a version. It should be a 'pointer' cursor (same as on the deployments page).

robot9706 (approved):
Tested, seems ok! ✨

polaroi8d (approved):
LGTM (informal)

m8vago (approved):

PR #750, opened 2023-07-14
fix(crux): failing deployment token migration

m8vago:
Fixes an issue with the deployment token audit log migration.


polaroi8d (approved):

polaroi8d:
By the way missing the scope.

PR #751, opened 2023-07-17
fix(crux-ui): change rolling version deployment error to warning

AmorfEvo:
- Fixed change rolling version deployment error to warning.
- Fixed change already have preparing deployment with the same prefix error to warning.
- Fixed show the warning message at least 5 seconds (only in these events).
- Fixed storage home button opens storages page instead of registries.
- Fixed switch the order of arguments and commands in the common config section.


polaroi8d (approved):
LGTM (informal)

ngyradi:
works, tested locally

m8vago (approved):

PR #753, opened 2023-07-18
fix(dagent): use none network mode & add unit tests

nandor-magyar:
None network mode added a bridge network too, instead of the special `none` network.
Unused network modes are removed from the UI and database schema, network now can only be: none, bridge, host.

Edit 1: add deprecations.


polaroi8d (approved):

ngyradi:
None is deployed correctly, but the other network modes are deployed as bridge.

polaroi8d (approved):

PR #754, opened 2023-07-18
fix(crux-ui): project name filter

AmorfEvo:
- Fixed project name filter (slugify the project name to create a prefix).
- Fixed secret fields and buttons not aligned.
- Fixed secret buttons border.
- Fixed kubernetes resource config memory typo.
- Fixed versioned copyability  e2e test.
- Split slowest e2e tests.


m8vago (approved):

robot9706 (commented):
When there's no value for a secret and the row buttons are hidden, the discard and save buttons kind of look awkward. Maybe the rows buttons could be visible but disabled or the discard and save buttons should be icon buttons next to the "SECRETS" heading, like the reset button?

image

AmorfEvo:
Ok, the first suggestion has been implemented, so the row buttons are now visible even when disabled.

robot9706 (approved):

PR #762, opened 2023-07-24
fix(dagent): remove host Rule if no root domain is given

nandor-magyar:
In dagent if no enough config provided: no domain, nor env variable the Host rule is omitted.
If Expose is set but no prefix or domain is provided, it error is thrown by the deployment.


robot9706 (approved):
Looks ok, but can't test it on Windows.

polaroi8d (approved):

PR #765, opened 2023-07-25
feat(agent): dagent ipc health

robot9706:
Implemented a unix socket based health check for dagent.


golang/internal/health/server.go
```diff
@@ -0,0 +1,81 @@
...
+}
+
+func SetHealthGRPCStatus(connected bool) {
+	health.Connected = connected
+}
+
+func Serve(ctx context.Context) error {
+	_, err := os.Stat(socketPath)
+	if err == nil {
+		err = os.RemoveAll(socketPath)
```
nandor-magyar, line 53: The first error is swallowed and only the remove error is printed. Is it intentional/ fail-safe?

golang/internal/health/types.go
```diff
@@ -0,0 +1,10 @@
+package health
+
+const (
+	socketType = "unix"
+	socketPath = "/tmp/agenthealth.sock"
```
nandor-magyar, line 5: A better path would be: path.Join(os.TempDir(),"dyrectorio","agenthealth.sock"), pay attention to file existence privilege error handling too.

nandor-magyar (commented):
Minor details, look neat otherwise. :sunglasses:

nandor-magyar:
Some small tests would be nice.

golang/internal/grpc/grpc.go
```diff
@@ -275,22 +282,24 @@ func grpcLoop(
...
 		if err != nil {
-			log.Error().Stack().Err(err).Msg("Cannot receive stream")
 			grpcConn.Client = nil
+			health.SetHealthGRPCStatus(false)
+
+			if err == io.EOF {
+				log.Info().Msg("End of stream")
+			} else {
+				log.Error().Stack().Err(err).Msg("Cannot receive stream")
+				// TODO replace the line above with an error status code check and terminate dagent accordingly
```
polaroi8d, line 299: Do we still need this comment, @nandor-magyar? Is this comment from the `GitLab` repository? If yes, what is blocking us from solving this issue?

polaroi8d (approved):
Minor comment,otherwise informal LGTM.

golang/internal/grpc/grpc.go
nandor-magyar: I don't remember exactly, but if EOF means that the remote connection is closed then agent should exit as well, other errors should be investigated what is possible to get here.

nandor-magyar (approved):

golang/internal/grpc/grpc.go
nandor-magyar: I'll create an issue.

PR #770, opened 2023-07-27
fix(crane): volume size parse error handling

robot9706:
Fixed crane crashing when trying to parse an invalid volume size.
Added volume size regex to UI.


polaroi8d (approved):
LGTM

nandor-magyar (approved):

PR #772, opened 2023-07-28
feat(crux-ui): save view mode for projects, deployment state, version state pages

AmorfEvo:
- Persist project view mode into localStorage.
- Refactor: view mode handling with custom hook for each page (we use this now on Projects, deployment state and version state pages).
- Fixed: show warning for insecure WebSocket connection in the production environment.
- Fixed: make view mode saving fully compatible with SSR.


web/crux-ui/src/pages/projects.tsx
```diff
@@ -51,7 +51,18 @@ const ProjectsPage = (props: ProjectsPageProps) => {
   const [creating, setCreating] = useState(false)
   const submitRef = useRef<() => Promise<any>>()
 
-  const [viewMode, setViewMode] = useState<ViewMode>('tile')
+  const [viewMode, setViewMode] = useState<ViewMode>(() => {
```
robot9706, line 54: I think it'd be better to implement a custom hook which would manage the ViewMode as well as save it to local storage. It could be something like `usePersistedViewMode(name)`. And in this case the name would be 'projects' so later (if we need to) we could use it somewhere else too.

AmorfEvo: We got a new usePersistedViewMode custom hook for this purpose, which is capable of handling view modes specific to each page. 
We now use this on the Projects, deployment state, and version state pages as well.

web/crux-ui/src/hooks/use-persisted-view-mode.ts
```diff
@@ -0,0 +1,28 @@
...
+}): [ViewMode, (mode: ViewMode) => void] {
+  const [viewMode, setViewMode] = useState<ViewMode>(() => {
+    if (typeof window !== 'undefined') {
+      const storedViewMode = localStorage.getItem(`viewMode_${pageName}`) as ViewMode
+      return storedViewMode || initialViewMode
+    }
+    return initialViewMode
+  })
+
+  useEffect(() => {
```
robot9706, line 19: You could also just return a custom setViewMode function which calls setState and localStorage.setItem, but that's just nitpicking :)

Otherwise tested, looks good 🚀

robot9706 (approved):

robot9706:
@m8vago pointed out an interesting problem due to SSR :)

First goto "/projects" then change the view mode to list and reload the page, this results in an `Error: Text content does not match server-rendered HTML.` error since SSR does not have access to localStorage. SSR renders the tile view, while the client rendes the list view. I'm not really sure what could be done, a simple "fix" could be to just disable hydration errors for this page, but that might not be the best. Could you look into this?

m8vago (approved):

PR #776, opened 2023-08-02
chore: new export dyrectorio offline bundle

nandor-magyar:
Refactored the previous version, added a minimal and a full make target.

### DIFF
--- a/offline/importar.sh
+++ b/offline/importar.sh
@@ -1,0 +1,2 @@
+#!/bin/sh
+ls -1  ./*.tar | while read -r line; do cat $line | docker load; done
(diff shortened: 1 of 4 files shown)


offline/importar.sh
polaroi8d, line 1: `import-full.sh` whould be better no?

polaroi8d (changes requested):

offline/importar.sh
nandor-magyar: Yes, the old one is removed, new is now `import.sh`.

polaroi8d (approved):

PR #779, opened 2023-08-03
fix: container builder image priority

robot9706:
Added `PullIfNewer` image priority and set it as default for the container builder. This priority pulls the image if the digests mismatch.


polaroi8d (approved):

polaroi8d (approved):
LGTM (informal)

nandor-magyar (approved):
LGTM!

PR #781, opened 2023-08-07
feat(web): protected deployment

robot9706:
- Add "Protected" toggle to "Edit deployments"
- Move `startDeploy` logic to a custom hook
- Fix unable to delete versions
- Fix selecting instances to deploy not working


AmorfEvo (approved):
LGTM

polaroi8d (approved):
Tested, works like a charm 🤩

web/crux-ui/src/components/projects/versions/deployments/add-deployment-card.tsx
m8vago, line 1: Is there any diff between this and the other add-deployment-card ?

web/crux/src/app/deploy/interceptors/deploy.start.interceptor.ts
```diff
@@ -91,6 +91,30 @@ export default class DeployStartValidationInterceptor implements NestInterceptor
       })
     }
 
+    if (!deployment.protected) {
+      const {
+        query: { ignoreProtected },
+      } = req
+
+      if (!ignoreProtected || ignoreProtected === false) {
```
m8vago, line 99: Is there a case when ignoreProtected is `!ignoreProtected` is `true` but it equals `false` ?

web/crux/src/app/deploy/interceptors/deploy.start.interceptor.ts
```diff
@@ -91,6 +91,30 @@ export default class DeployStartValidationInterceptor implements NestInterceptor
...
+        const otherProtected = await this.prisma.deployment.findFirst({
+          where: {
+            protected: true,
+            nodeId: deployment.nodeId,
+            prefix: deployment.prefix,
+          },
+        })
+
+        if (otherProtected != null) {
```
m8vago, line 108: I think this warning should be thrown when:
Rolling versions: `There is another deployment, which is protected on the same node same prefix`
Incremental: `There is another deployment, which is protected on the same node same prefix and it isn't this version's parent or child.`

m8vago (changes requested):

web/crux-ui/src/components/projects/versions/deployments/add-deployment-card.tsx
robot9706: The other has the project and version selectors and some extra logic for them. But in terms of protected deployment stuff they are the same.

web/crux/src/app/deploy/interceptors/deploy.start.interceptor.ts
robot9706: I wanted to handle the case when `ignoreProtected` is null, but yeah they are the same. I'll remove the second condition.

web/crux/src/app/deploy/interceptors/deploy.start.interceptor.ts
robot9706: So in case of an incremental deployment I should also check if there's another deployment which is in another version.

web/crux/src/app/deploy/interceptors/deploy.create.interceptor.ts
```diff
@@ -62,6 +62,33 @@ export default class DeployCreateValidationInterceptor implements NestIntercepto
...
+          versionId:
+            version.type === 'incremental'
+              ? {
+                  not: body.versionId,
+                }
+              : undefined,
+        },
+      })
+
+      if (otherProtected != null) {
```
m8vago, line 80: `if (otherProtected)` would be nice.

web/crux/src/app/deploy/interceptors/deploy.patch.interceptor.ts
```diff
@@ -63,6 +63,33 @@ export default class DeployPatchValidationInterceptor implements NestInterceptor
...
+          versionId:
+            deployment.version.type === 'incremental'
+              ? {
+                  not: deployment.versionId,
+                }
+              : undefined,
+        },
+      })
+
+      if (otherProtected != null) {
```
m8vago, line 81: same as above.

web/crux/src/app/deploy/interceptors/deploy.start.interceptor.ts
m8vago: Yep. The query now looks good!

web/crux/src/app/deploy/interceptors/deploy.start.interceptor.ts
```diff
@@ -91,6 +91,39 @@ export default class DeployStartValidationInterceptor implements NestInterceptor
...
+            versionId:
+              deployment.version.type === 'incremental'
+                ? {
+                    not: deployment.versionId,
+                  }
+                : undefined,
+          },
+        })
+
+        if (otherProtected != null) {
```
m8vago, line 114: Simply `if (otherProtected)`

m8vago (approved):
Just some minor inconvenience in ifs. We can fix that in a later pr.

m8vago (approved):

PR #784, opened 2023-08-11
fix(crux-ui): register and team card validation

ngyradi:
Add validation to team slug and register formik.


web/crux-ui/src/validations/auth.ts
```diff
@@ -1,22 +1,42 @@
...
-
-export const registerSchema = yup.object().when('method', {
-  is: 'password',
-  then: () => registerWithPasswordSchema,
-  otherwise: () => registerWithOidcSchema,
+  firstName: identityNameRule.required(),
+  lastName: identityNameRule.required(),
+})
+
+export const registerSchema = yup.object().shape({
```
robot9706, line 18: This is unused, no?

ngyradi: It's used in api register.ts.

polaroi8d (approved):
LGTM (informal)

robot9706 (approved):

web/crux-ui/src/pages/auth/register.tsx
```diff
@@ -197,6 +197,12 @@ const RegisterPage = (props: RegisterPageProps) => {
               type="password"
               onChange={formik.handleChange}
               value={formik.values.confirmPassword}
+            />
+            <DyoMessage
```
m8vago, line 201: please spread the html elements next time, to improve readability.

web/crux-ui/src/validations/auth.ts
```diff
@@ -1,22 +1,43 @@
...
+  firstName: identityNameRule.required(),
+  lastName: identityNameRule.min(0),
+  email: yup.string().email().required(),
 })
 
-export const registerSchema = yup.object().when('method', {
-  is: 'password',
-  then: () => registerWithPasswordSchema,
-  otherwise: () => registerWithOidcSchema,
+export const registerSchema = yup.object().shape({
```
m8vago, line 19: What was the problem with the original solution?

m8vago (changes requested):

web/crux-ui/src/validations/auth.ts
ngyradi: It returned a yup object with empty fields so it didn't validate anything.

robot9706: yup can only validate fields which is in the schema and since 'method' is not part of the schema its value is always undefined.

m8vago (approved):

PR #788, opened 2023-08-14
feat(web): config bundles

robot9706:
- Added ConfigBundles sidebar menu option
- Added DyoMultiSelect component
- Added ConfigBundle selector to deployment (above environment)

Notes: 
- When deploying all config bundle keys are tested to see if any of them collide with eachother which results in an error message (deployment environments override these).
- On a successful incremental deployment all merged environments are saved into the deployment and config bundles are removed to see what was exactly deployed.

Depends on: #781


AmorfEvo (approved):
LGTM

polaroi8d:
If a deployment is associated with only one bundle, and a user wishes to view its details, kindly direct the user to the configuration bundle deal page rather than the configuration bundle list.

There are a few validation points that appear to be missing, but we can discuss those in person.

web/crux-ui/locales/en/config-bundles.json
```diff
@@ -0,0 +1,10 @@
+{
+    "configBundleName": "Config bundle - {{name}}",
+    "name": "Name",
+    "view": "View {{name}}",
+    "new": "New config bundle",
+    "noItems": "You haven't added a config bundle. To add a new one, click 'Add' on the top right.",
+    "environment": "Environment",
+    "tips": "Config bundles are environments which can be assigned to deployments.",
+    "inUse": "This config bundle is already in use in some deployments."
+}
```
polaroi8d, line 10: No NL EOF

web/crux-ui/public/configbundles.svg
polaroi8d, line 1: Why we need 2 svg with the same content with different name?

web/crux-ui/src/components/projects/versions/images/config/common-config-section.tsx
```diff
@@ -457,7 +457,7 @@ const CommonConfigSection = (props: CommonConfigSectionProps) => {
                   className="w-full ml-2"
                   choices={storages ? [null, ...storages.map(it => it.id)] : [null]}
                   selection={config.storage?.storageId ?? null}
-                  converter={(it: string) => storages?.find(storage => storage.id === it)?.name ?? t('common.none')}
+                  converter={(it: string) => storages?.find(storage => storage.id === it)?.name ?? t('common:none')}
```
polaroi8d, line 460: 🦅

web/crux-ui/src/const.ts
```diff
@@ -6,6 +6,7 @@ export const IMAGE_FILTER_MIN_LENGTH = 1 // characters
 export const IMAGE_WS_REQUEST_DELAY = 500 // millis
```
polaroi8d, line 6: In this context, we are currently employing the same approach. Would it be more advantageous to utilize a single common approach? Alternatively, do you foresee any significant differences arising between these methods in the future?

```
export const IMAGE_WS_REQUEST_DELAY = 500 // millis
export const INSTANCE_WS_REQUEST_DELAY = IMAGE_WS_REQUEST_DELAY // millis
export const DEPLOYMENT_EDIT_WS_REQUEST_DELAY = 500 // millis
export const CONFIG_BUNDLE_EDIT_WS_REQUEST_DELAY = 500 // millis```

File: web/crux-ui/src/routes.ts [hunk 4]
```diff
@@ -531,6 +531,46 @@ class StorageRoutes {
   details = (id: string) => `${this.root}/${id}`
 }
 
+// config bundle
+
+class ConfigBundleApi {
+  private readonly root: string
+
+  constructor(root: string) {
+    this.root = `/api${root}`
```
polaroi8d, line 540: Is this good?  `/api${root}`

web/crux-ui/src/routes.ts
```diff
@@ -531,6 +531,46 @@ class StorageRoutes {
...
+  details = (id: string) => `${this.root}/${id}`
+
+  options = () => `${this.root}/options`
+}
+
+class ConfigBundleRoutes {
+  private readonly root: string
+
+  constructor(root: string) {
+    this.root = `${root}/configbundles`
```
polaroi8d, line 554: I'm thinking on it, which would be more proper this, or with the -, like `config-bundles`

web/crux/prisma/migrations/20230810111242_add_config_bundles/migration.sql
```diff
@@ -0,0 +1,37 @@
...
+ALTER TABLE "ConfigBundle" ADD CONSTRAINT "ConfigBundle_teamId_fkey" FOREIGN KEY ("teamId") REFERENCES "Team"("id") ON DELETE CASCADE ON UPDATE CASCADE;
+
+-- AddForeignKey
+ALTER TABLE "ConfigBundleOnDeployments" ADD CONSTRAINT "ConfigBundleOnDeployments_deploymentId_fkey" FOREIGN KEY ("deploymentId") REFERENCES "Deployment"("id") ON DELETE CASCADE ON UPDATE CASCADE;
+
+-- AddForeignKey
+ALTER TABLE "ConfigBundleOnDeployments" ADD CONSTRAINT "ConfigBundleOnDeployments_configBundleId_fkey" FOREIGN KEY ("configBundleId") REFERENCES "ConfigBundle"("id") ON DELETE RESTRICT ON UPDATE CASCADE;
+
+-- AddForeignKey
+ALTER TABLE "ConfigBundleOnDeployments" ADD CONSTRAINT "ConfigBundleOnDeployments_teamId_fkey" FOREIGN KEY ("teamId") REFERENCES "Team"("id") ON DELETE CASCADE ON UPDATE CASCADE;
```
polaroi8d, line 37: NL

web/crux/prisma/schema.prisma
```diff
@@ -525,3 +528,33 @@ model Storage {
 
   @@unique([name, teamId])
 }
+
+model ConfigBundle {
+  id        String   @id @default(uuid()) @db.Uuid
+  name      String
```
polaroi8d, line 534: Didn't we want to include a brief description for these?

web/crux/src/app/config.bundle/config.bundle.message.ts
```diff
@@ -0,0 +1,15 @@
+import { UniqueKeyValue } from 'src/domain/container'
+
+export const WS_TYPE_PATCH_CONFIG_BUNDLE = 'patch-config-bundle'
+export type PatchConfigBundleEnvMessage = {
+  name?: string
+  environment?: UniqueKeyValue[]
+}
+
+export const WS_TYPE_CONFIG_BUNDLE_UPDATED = 'config-bundle-updated'
+export type ConfigBundleEnvUpdatedMessage = {
```
polaroi8d, line 10: Regarding that the 2 object content is the same, it would be better to use 1 outsourced, no? 🤔

web/crux/src/app/deploy/deploy.service.ts
```diff
@@ -886,6 +954,30 @@ export default class DeployService {
...
+
+    return [me, editors.getEditors()]
+  }
+
+  async onEditorLeft(deploymentId: string, clientToken: string): Promise<EditorLeftMessage> {
+    const editors = await this.editorServices.getOrCreateService(deploymentId)
+    const message = editors.onClientLeft(clientToken)
+
+    if (editors.editorCount < 1) {
+      this.logger.verbose(`All editors left removing ${deploymentId}`)
```
polaroi8d, line 974: What is the use case for this? I'm missing an understanding of when this situation might arise.

web/crux/src/app/deploy/interceptors/deploy.start.interceptor.ts
```diff
@@ -124,6 +129,34 @@ export default class DeployStartValidationInterceptor implements NestInterceptor
       }
     }
 
+    if (deployment.configBundles.length > 0) {
+      const deploymentEnv = (deployment.environment as UniqueKeyValue[]) ?? []
+      const deploymentEnvKeys = deploymentEnv.map(it => it.key)
+
+      const seenKeys: Record<string, string> = {} // Environment key -> config bundle name
```
polaroi8d, line 136: Do not really understand the comment in here. Could you please elaborate a bit more? I can't came up with a better naming but `seenKeys` a bit shady for me 🍨

polaroi8d (changes requested):
Minor comments and questions, otherhand good job! 🧨

web/crux/src/app/deploy/interceptors/deploy.start.interceptor.ts
AmorfEvo: What about checkedKeys?

polaroi8d: Better for me

web/crux-ui/public/configbundles.svg
robot9706: I'm not sure, but this is how it was done with registries and I based the config bundle UI on that. I can use one svg though.

web/crux-ui/src/const.ts
robot9706: I prefer having separate consts even though they are the same, but if it looks better I could merge them into something like `ENV_EDITOR_THROTTLE` or `EDITOR_WS_REQUEST_DELAY` or something like that.

polaroi8d: Okay we can leave it, just don't mix the usage like here: `INSTANCE_WS_REQUEST_DELAY = IMAGE_WS_REQUEST_DELAY`, IMHO.

web/crux-ui/src/routes.ts
robot9706: It should be OK, because root comes from TeamRoutes which is `/teamSlug`, so it'll be `/api/teamSlug/....`.

web/crux-ui/public/configbundles.svg
polaroi8d: Could you please review the registries? If there's only one necessary, kindly delete the others.

web/crux-ui/src/routes.ts
polaroi8d: Ahh, okay just looks weird a little, resolved my comment.

web/crux/src/app/deploy/deploy.service.ts
robot9706: For example two users open the same deployment to edit, then the editor service will have 2 editors for the same deployment, then both of them close their browsers, this way the deployment will have no editors, so we'll free the underlying services (which hold info about focused editors, editor names, colors, etc.).

web/crux/src/app/deploy/interceptors/deploy.start.interceptor.ts
robot9706: With the comment I wanted to suggest that it's a mapping which connects environment keys to config bundle names. Maybe using `envToBundle` would be a better variable name, and I could then remove the comment?

web/crux/src/app/config.bundle/config.bundle.message.ts
robot9706: We have separate types for each message type even if they are the same, so I just wanted to follow how we do it generally.

Also I think having different types is beneficial, because the type name suggests what it does, instead of having something like `DeploymentEnvironmentMessage`.

web/crux/prisma/schema.prisma
robot9706: I'm not sure, but I can add a simple description field easily.

polaroi8d: Yeah, please.

web/crux/src/app/config.bundle/config.bundle.message.ts
polaroi8d: Okey, it was just a question, resolving the comment.

web/crux/src/app/deploy/deploy.service.ts
polaroi8d: Okey, thanks.

web/crux/src/app/deploy/interceptors/deploy.start.interceptor.ts
polaroi8d: @robot9706 good idea, yea.

web/crux-ui/public/configbundles.svg
polaroi8d: Resolved, thanks!

web/crux-ui/src/pages/[teamSlug]/config-bundles/[configBundleId].tsx
```diff
@@ -0,0 +1,157 @@
...
+
+const ConfigBundleDetailsPage = (props: ConfigBundleDetailsPageProps) => {
+  const { configBundle: propsConfigBundle } = props
+
+  const { t } = useTranslation('config-bundles')
+  const routes = useTeamRoutes()
+
+  const onWsError = (error: Error) => {
+    // eslint-disable-next-line
+    console.error('ws', 'edit-deployment', error)
```
Comment (m8vago) [inline web/crux-ui/src/pages/[teamSlug]/config-bundles/[configBundleId].tsx:35]: edit-config-bundle?

web/crux/prisma/schema.prisma
```diff
@@ -525,3 +528,34 @@ model Storage {
...
+  deployments ConfigBundleOnDeployments[]
+
+  @@unique([name, teamId])
+}
+
+model ConfigBundleOnDeployments {
+  deployment   Deployment @relation(fields: [deploymentId], references: [id], onDelete: Cascade)
+  deploymentId String     @db.Uuid
+
+  configBundle   ConfigBundle @relation(fields: [configBundleId], references: [id])
```
m8vago, line 554: Let's check if set this  to cascade, then we can probably get rid of the team relation.

web/crux/src/interceptors/prisma-error-interceptor.ts
```diff
@@ -60,7 +60,7 @@ export default class PrismaErrorInterceptor implements NestInterceptor {
 
     if (err.code === UUID_INVALID) {
       return new CruxBadRequestException({
-        message: 'Invalid uuid',
+        message: `UUID_INVALID ${err.message}, ${err.stack ?? 'no stack'}, ${JSON.stringify(err.meta)}`,
```
m8vago, line 63: How come this message?

m8vago (changes requested):

web/crux/prisma/schema.prisma
robot9706: Seems to be working with cascade without teamId :)

m8vago (approved):

polaroi8d (approved):

PR #791, opened 2023-08-16
feat(crux-ui): image list in registry page

ngyradi:
Added a list for images in a registry.


web/crux-ui/e2e/with-login/registry.spec.ts
```diff
@@ -93,3 +93,30 @@ test("Unchecked registry shouldn't search images", async ({ page }) => {
...
+
+  await page.locator('text=Add').click()
+  await expect(page.locator('h4')).toContainText('New registry')
+
+  await page.locator('input[name=name] >> visible=true').fill(registryName)
+  await page.locator('form >> text=Docker Hub').click()
+  await expect(await page.locator('label[for=imageNamePrefix]')).toContainText('Organization name or username')
+  await page.locator('input[name=imageNamePrefix]').fill('library')
+
+  await page.screenshot({ path: screenshotPath('registry_new_filled'), fullPage: true })
```
robot9706, line 110: This screenshot name is also used in the same file elsewhere which would overwrite it. I think you could remove the screenshot because we already have one from the create registry test.

web/crux-ui/src/pages/[teamSlug]/registries/[registryId].tsx
```diff
@@ -54,6 +121,43 @@ const RegistryDetailsPage = (props: RegistryDetailsPageProps) => {
...
+
+  useEffect(() => {
+    sock.send(WS_TYPE_FIND_IMAGE, {
+      registryId: registry.id,
+      filter: '',
+    } as FindImageMessage)
+  }, [])
+
+  useEffect(() => {
+    filters.setItems(images)
```
Comment (robot9706) [inline web/crux-ui/src/pages/[teamSlug]/registries/[registryId].tsx:153]: You could move this to the socket WS_TYPE_FIND_IMAGE_RESULT message handler.

web/crux-ui/src/models/common.ts
```diff
@@ -42,6 +42,11 @@ export type PaginationQuery = {
   to: string
 }
 
+// sorting
+
+export const SORTING_TYPE_VALUES = ['none', 'az', 'za'] as const
```
m8vago, line 47: `asc` and `desc` would be better IMHO, instead of `az` and `za`

m8vago (approved):

polaroi8d (changes requested):
First of all, when opening a registry, the message that there is no image appears slightly different to me. On the other hand, we are not utilizing the ascending / descending options in the filter box. Instead, we have implemented functionality similar to the deployment screen: when users click on a column header, they can sort the respective column. 

Talk about the possible solutions in person.

robot9706:
@polaroi8d Just refactored the PR a bit, removed the ascending/descending buttons and the table column sorts instead, as well as added a loading indicator.

m8vago (approved):

polaroi8d (approved):
Thanks! LGTM

robot9706 (approved):

PR #792, opened 2023-08-16
feat(crux): add mlops template

AmorfEvo:
- Add the MLFlow template to the list.
- Fix: port config rule validation.


web/crux-ui/src/validations/container.ts
```diff
@@ -207,7 +207,8 @@ const createOverlapTest = (
   )
 
 const portConfigRule = yup.mixed().when('portRanges', portRanges => {
-  if (!portRanges) {
+  // TODO(amorfevo): quickfixed, but why are we able to come here with such strange values at all?
+  if (!portRanges || !portRanges[0] || portRanges[0].length === 0) {
```
polaroi8d, line 211: `!portRanges?.[0]?.length` but summon js ninjas here @robot9706 @m8vago

AmorfEvo: Yes, it's shorter and the same. I apply it.

robot9706: Or it could also be `!portRanges || Array.isArray(portRanges)` since it a portRange can't be an array.

polaroi8d: @robot9706 love that one, more readable

web/crux-ui/src/validations/container.ts
```diff
@@ -207,7 +207,8 @@ const createOverlapTest = (
   )
 
 const portConfigRule = yup.mixed().when('portRanges', portRanges => {
```
robot9706, line 209: It seems to me that the actual fix is to just wrap the portRanges in an array like `([portRanges]) => ...`.
Because yup passes all references fields to the function, in this case there's only one reference but it still seems to pass it as an array.

AmorfEvo: Well, this is the cleanest.

robot9706 (approved):

PR #796, opened 2023-08-23
feat: oneshot agent install tokens

m8vago:
Rework agent connection handling. From now on there are two types of tokens. 
- Install tokens: Agents receive this as the `GRPC_TOKEN` env. They exchange it on the first run to a connection token.
- Connection tokens: These are the ones, which are used for regular connections.

During an update, agents gets a new token.

Also there is a workaround for legacy agents, whom are getting shutdown and removed by the newly updated agent, so they do not stuck in a crashloop.


golang/cmd/crane/crane.go
```diff
@@ -46,41 +45,44 @@ func main() {
...
+	cfg.InjectPrivateKey(secretHandler)
 	if err != nil {
-		return nil, err
+		log.Panic().Err(err).Msg("Failed to load private key")
 	}
 
-	commonConfig.InjectSecret(secret, &cfg.CommonConfiguration)
+	err = cfg.InjectGrpcToken(secretHandler)
+	if err != nil {
+		log.Panic().Err(err).Msg("Failed to load grpc token")
```
polaroi8d, line 66: gRPC

golang/internal/config/secret.go
```diff
@@ -111,6 +73,56 @@ func IsExpiredKey(fileContent string) (bool, error) {
...
+	if err == nil && token != "" {
+		c.GrpcToken, err = ValidateJwtAndCheckNonceBlacklist(secrets, token)
+		if err == nil {
+			return nil
+		}
+	}
+
+	token = c.InstallToken
+	if err != nil {
+		log.Error().Err(err).Msg("Failed to validate the connection token. Failing back to the install token")
```
polaroi8d, line 115: @nandor-magyar always tells me that we are using sentences with small capital starters

golang/internal/config/secret.go
```diff
@@ -111,6 +73,56 @@ func IsExpiredKey(fileContent string) (bool, error) {
...
+		return err
+	}
+
+	c.SecretPrivateKey = key
+	return nil
+}
+
+func (c *CommonConfiguration) InjectGrpcToken(secrets SecretStore) error {
+	token, err := secrets.GetConnectionToken()
+	if err == nil && token != "" {
```
polaroi8d, line 106: Unnecessary double check.

```
if token != "" {
    c.GrpcToken, err = ValidateJwtAndCheckNonceBlacklist(secrets, token)
    if err == nil {
        return nil
    }
}
```

golang/internal/config/secret.go
```diff
@@ -111,6 +73,56 @@ func IsExpiredKey(fileContent string) (bool, error) {
...
+		}
+	}
+
+	token = c.InstallToken
+	if err != nil {
+		log.Error().Err(err).Msg("Failed to validate the connection token. Failing back to the install token")
+	}
+
+	if token == "" {
+		return fmt.Errorf("no token provided")
```
polaroi8d, line 119: Use `log.Error()` please.

golang/internal/grpc/grpc.go
```diff
@@ -62,16 +65,30 @@ type ContainerWatchContext struct {
...
+type UpdateOptions struct {
+	UpdateAlways  bool
+	UseContainers bool
+}
+
+type ClientLoop struct {
+	Ctx         context.Context
+	WorkerFuncs WorkerFunctions
+	Secrets     config.SecretStore
+	cancel      context.CancelFunc
```
polaroi8d, line 77: Cancel?

golang/internal/grpc/grpc.go
```diff
@@ -88,7 +105,10 @@ type WorkerFunctions struct {
 
 type contextKey int
 
-const contextConfigKey contextKey = 0
+const (
+	contextConfigKey     contextKey = 0
+	contextMetadataToken            = "dyo-node-token"
```
polaroi8d, line 110: Is this okay to miss the string?

golang/internal/grpc/grpc.go
```diff
@@ -150,26 +170,34 @@ func Init(grpcContext context.Context,
...
+	secrets config.SecretStore,
 ) {
 	log.Info().Msg("Spinning up gRPC Agent client...")
 	if grpcConn == nil {
 		grpcConn = &Connection{}
 	}
 
 	ctx, cancel := context.WithCancel(grpcContext)
+	loop := ClientLoop{
+		cancel:      cancel,
```
polaroi8d, line 182: Cancel

golang/internal/grpc/grpc.go
```diff
@@ -214,65 +242,57 @@ func Init(grpcContext context.Context,
...
 }
 
-func grpcLoop(
-	ctx context.Context,
-	connParams *ConnectionParams,
-	workerFuncs WorkerFunctions,
-	cancel context.CancelFunc,
-	appConfig *config.CommonConfiguration,
-) {
+func (l *ClientLoop) grpcLoop(connParams *ConnectionParams) {
```
polaroi8d, line 278: Minor: cl?

golang/internal/grpc/grpc.go
```diff
@@ -289,6 +309,33 @@ func grpcLoop(
 		command := new(agent.AgentCommand)
 		err = stream.RecvMsg(command)
 		if err != nil {
+			s := status.Convert(err)
+			if s != nil && (s.Code() == codes.Unauthenticated || s.Code() == codes.PermissionDenied || s.Code() == codes.NotFound) {
+				if l.AppConfig.GrpcToken.Type == config.Connection {
+					log.Error().Err(err).Msg("Invalid connection token. Removing")
```
polaroi8d, line 315: Small capital.

golang/internal/grpc/grpc.go
```diff
@@ -289,6 +309,33 @@ func grpcLoop(
...
+
+					// overwrite jwt token
+					err := l.Secrets.SaveConnectionToken("")
+					if err != nil {
+						log.Err(err).Msg("Failed to delete the invalid connection token. Exiting")
```
polaroi8d, line 320: Small capital letter

golang/internal/grpc/grpc.go
```diff
@@ -289,6 +309,33 @@ func grpcLoop(
...
+
+						os.Exit(1)
```
polaroi8d, line 322: Missing the debug message like below.

golang/internal/grpc/grpc.go
```diff
@@ -591,41 +638,65 @@ func executeSecretList(
...
 	if updateFunc == nil {
 		log.Error().Msg("Self update function not implemented")
 		return
 	}
 
-	err := updateFunc(ctx, command.Tag, command.TimeoutSeconds)
+	_, err := config.ValidateAndCreateJWT(command.Token)
 	if err != nil {
-		log.Error().Stack().Err(err).Msg("Update error")
+		log.Error().Msg("Jwt validation failed")
```
polaroi8d, line 649: JWT or jwt

golang/internal/grpc/grpc.go
```diff
@@ -591,41 +638,65 @@ func executeSecretList(
...
 
-		_, err := grpcConn.Client.AbortUpdate(ctx, resp)
-		if err != nil {
-			log.Error().Stack().Err(err).Msg("Update abort request error")
-		}
+	err = loop.executeReplaceToken(&agent.ReplaceTokenRequest{
+		Token: command.Token,
+	})
+
+	if err == nil {
```
polaroi8d, line 659: We are not using onelines for error handling? @nandor-magyar 
```
if err := updateFunc(loop.Ctx, command, updateOptionsFromAppConfig(loop.AppConfig)); err != nil {
	abortUpdate(loop.Ctx, err)
}
```

golang/internal/grpc/grpc.go
```diff
@@ -754,6 +825,53 @@ func executeContainerLog(ctx context.Context, command *agent.ContainerLogRequest
...
+
+	_, err = grpcConn.Client.TokenReplaced(l.Ctx, &common.Empty{})
+	if err != nil {
+		log.Error().Err(err).Msg("Failed to report token replacement, falling back to the old token")
+		return err
+	}
+
+	err = l.Secrets.SaveConnectionToken(command.GetToken())
+	if err != nil {
+		// For example, out of space?
```
polaroi8d, line 845: Please provide an orderly comment if you want to provide some note for the code. (`// NOTE():`)

golang/pkg/dagent/config/secret.go
```diff
@@ -1,26 +1,120 @@
...
+	}
+
+	if errors.Is(err, os.ErrNotExist) {
+		log.Debug().Msgf("Key file does not exist: %v", path)
+		return generateKey(path)
+	} else if err != nil {
+		return "", fmt.Errorf("key file can't be read: %w", err)
+	}
+
+	// exists but expired -> migrate present keys?!
```
polaroi8d, line 36: Please elaborate on this comment.

web/crux-ui/src/components/nodes/node-status-indicator.tsx
```diff
@@ -9,7 +9,7 @@ const statusToColor = (status: NodeStatus) => {
     case 'connected':
       return 'bg-dyo-green'
     case 'outdated':
-      return 'bg-dyo-purple'
+      return 'bg-dyo-violet'
```
polaroi8d, line 12: Why?

web/crux/prisma/migrations/20230807153730_oneshot_agent_token/migration.sql
```diff
@@ -0,0 +1,68 @@
...
+    CONSTRAINT "NodeEvent_pkey" PRIMARY KEY ("id")
+);
+
+-- AddForeignKey
+ALTER TABLE "NodeEvent" ADD CONSTRAINT "NodeEvent_nodeId_fkey" FOREIGN KEY ("nodeId") REFERENCES "Node"("id") ON DELETE CASCADE ON UPDATE CASCADE;
+
+insert into "NodeEvent"
+select * from "_prisma_migrations_AgentEvent";
+
+drop table "_prisma_migrations_AgentEvent";
```
polaroi8d, line 68: Is this gonna work with production rollout? 🤔

web/crux/src/app/agent/connection-strategies/agent.connection.strategy.ts
```diff
@@ -0,0 +1,118 @@
...
+    })
+  }
+
+  protected parseToken(connection: GrpcNodeConnection, request: AgentInfo): AgentToken {
+    const token = this.jwtService.decode(connection.jwt) as AgentToken
+
+    const nodeId = token.sub
+    if (request.id !== nodeId) {
+      throw new CruxBadRequestException({
+        message: 'Id mismatch.',
```
polaroi8d, line 57: Node ID mismatch, no?

web/crux/src/shared/const.ts
```diff
@@ -11,10 +11,11 @@ export const VERSIONLESS_PROJECT_VERSION_NAME = 'rolling'
 
 const DAY_IN_MILLIS = 24 * 60 * 60 * 1000
 export const TEAM_INVITATION_EXPIRATION = 7 * DAY_IN_MILLIS
-export const JWT_EXPIRATION = 10 * 60 * 1000 // 10 minutes
+export const JWT_EXPIRATION_MILLIS = 10 * 60 * 1000 // 10 minutes
 export const CONTAINER_DELETE_TIMEOUT = 1000 // millis
 
 export const DEFAULT_CONTAINER_LOG_TAIL = 40
+export const MINIMUM_AGENT_VERSION = '0.7.0'
```
polaroi8d, line 18: We have to always update this with committing new versions?

polaroi8d: ...by the way, where is the usage for this const?

polaroi8d (changes requested):
Not tested yet, add some questions and reviews. I started commenting on all Golang error handling but after a few I stopped, we start sentences with lowercase letters.

golang/internal/config/secret.go
nandor-magyar: That is true for errors, that is a string.

polaroi8d:
When I attempted to update my currently running agent, I encountered a segmentation violation.

```
{"level":"debug","containerID":"15459d1e3593cbc01dde6e65fee6895589a71e156dc3ad8cba9215d8d88c1aa9","time":"2023-08-24T13:52:25Z","message":"Created new dagent"}
panic: runtime error: invalid memory address or nil pointer dereference
[signal SIGSEGV: segmentation violation code=0x1 addr=0x18 pc=0xe9a3ae]

goroutine 290 [running]:
github.com/dyrector-io/dyrectorio/golang/pkg/dagent/update.RewriteUpdateErrors({0x0, 0x0})
	/__w/dyrectorio/dyrectorio/golang/pkg/dagent/update/update.go:214 +0x8e
github.com/dyrector-io/dyrectorio/golang/pkg/dagent/update.SelfUpdate({0x126fa70, 0xc0004215f0}, {0xc0007201d0, 0x6}, 0xa?)
	/__w/dyrectorio/dyrectorio/golang/pkg/dagent/update/update.go:116 +0x111
github.com/dyrector-io/dyrectorio/golang/internal/grpc.executeUpdate({0x126fa70, 0xc0004215f0}, 0xc00032e000?, 0xc0009022a0?)
	/__w/dyrectorio/dyrectorio/golang/internal/grpc/grpc.go:591 +0x69
created by github.com/dyrector-io/dyrectorio/golang/internal/grpc.grpcProcessCommand
	/__w/dyrectorio/dyrectorio/golang/internal/grpc/grpc.go:231 +0x66a
```

golang/cmd/crane/crane.go
m8vago: I think here the camel case is more dominant. @nandor-magyar ?

polaroi8d: image
proof

golang/cmd/dagent/.env.example
```diff
@@ -32,4 +34,3 @@ TRAEFIK_LOG_LEVEL=
 TRAEFIK_TLS=false
 DEFAULT_REGISTRY=index.docker.io
 WEBHOOK_TOKEN=
-SECRET_PRIVATE_KEY_FILE=/path/to/secret.key
```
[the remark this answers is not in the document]
m8vago: We don't need that in an env. Discussed it with Nandor. Basically we use the internal mount path variable to determine the folder and append the constant file names.

golang/internal/config/secret.go
```diff
@@ -4,48 +4,25 @@ File for all the secret key file parsing and initializiation
...
-		return "", keyErr
-	}
-
-	if privateKeyObj == nil {
-		return "", fmt.Errorf("key file is nil: %v", secretPath)
-	}
-
-	if !privateKeyObj.IsExpired() {
-		keyStr, keyErr := privateKeyObj.ArmorWithCustomHeaders("", "")
+func (m *NonceBlacklistedError) Error() string {
```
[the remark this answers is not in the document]
m8vago: So we can log the error.

golang/cmd/crane/crane.go
m8vago: I can see that, I'm not saying this isn't how google prefers to write grpc. I am saying this should not affect the function names, as coding conventions has higher priority than marketing ones.

golang/internal/config/secret.go
m8vago: That's not the same thing. This returns an error instance, which is going get logged in the caller function.

golang/internal/grpc/grpc.go
m8vago: That's for canceling the grpc call, thus the loop itself. What would you prefer?

golang/internal/grpc/grpc.go
m8vago: @nandor-magyar ?

golang/internal/grpc/grpc.go
m8vago: As mentioned above, that is true for error strings only.

golang/internal/grpc/grpc.go
m8vago: It is present on the error log up above.

golang/cmd/crane/crane.go
nandor-magyar: The comment is against a string, not a variable name, in this case gRPC it is, otherwise in function names, yes. Once again, I'm 100% sure this was **directed** towards the message.

web/crux/src/shared/const.ts
m8vago: This is not ready yet. I will implement that in a new ticket.

web/crux/src/interceptors/prisma-error-interceptor.ts
```diff
@@ -103,6 +103,7 @@ export default class PrismaErrorInterceptor implements NestInterceptor {
     NotificationEvent: 'notificationEvent',
     Token: 'token',
     Storage: 'storage',
-    AgentEvent: 'agentEvent',
+    NodeEvent: 'nodeEvent',
+    NodeToken: 'token',
```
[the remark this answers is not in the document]
m8vago: This is used for translating a psql not found error to the user (in this case the agent), so i believe we can safely write token in that context.

golang/cmd/crane/crane.go
m8vago: Ah, silly me. I didn't noticed that. Sorry!

web/crux/prisma/migrations/20230807153730_oneshot_agent_token/migration.sql
polaroi8d: It was a "koltoi kerdes" 🤡

web/crux/src/interceptors/prisma-error-interceptor.ts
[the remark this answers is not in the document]
polaroi8d: Okey, just confusing a little bit, at least to me.

golang/internal/config/secret.go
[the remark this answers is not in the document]
m8vago: @nandor-magyar is there a more preferred way to define a custom error?

golang/internal/grpc/grpc.go
m8vago: or the type definition?

web/crux/src/interceptors/prisma-error-interceptor.ts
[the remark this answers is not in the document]
m8vago: I see. I'll make a comment then.

golang/internal/grpc/grpc.go
nandor-magyar: It is okay as it is for constants.

golang/internal/grpc/grpc.go
nandor-magyar: If there err is already defined, a shadow warning is shown, that would make the lint fail, so it is fine. I don't we need a rule for this, both forms are correct, whichever is more readable is to be used.

m8vago:
> When I attempted to update my currently running agent, I encountered a segmentation violation.
> 
> ```
> {"level":"debug","containerID":"15459d1e3593cbc01dde6e65fee6895589a71e156dc3ad8cba9215d8d88c1aa9","time":"2023-08-24T13:52:25Z","message":"Created new dagent"}
> panic: runtime error: invalid memory address or nil pointer dereference
> [signal SIGSEGV: segmentation violation code=0x1 addr=0x18 pc=0xe9a3ae]
> 
> goroutine 290 [running]:
> github.com/dyrector-io/dyrectorio/golang/pkg/dagent/update.RewriteUpdateErrors({0x0, 0x0})
> 	/__w/dyrectorio/dyrectorio/golang/pkg/dagent/update/update.go:214 +0x8e
> github.com/dyrector-io/dyrectorio/golang/pkg/dagent/update.SelfUpdate({0x126fa70, 0xc0004215f0}, {0xc0007201d0, 0x6}, 0xa?)
> 	/__w/dyrectorio/dyrectorio/golang/pkg/dagent/update/update.go:116 +0x111
> github.com/dyrector-io/dyrectorio/golang/internal/grpc.executeUpdate({0x126fa70, 0xc0004215f0}, 0xc00032e000?, 0xc0009022a0?)
> 	/__w/dyrectorio/dyrectorio/golang/internal/grpc/grpc.go:591 +0x69
> created by github.com/dyrector-io/dyrectorio/golang/internal/grpc.grpcProcessCommand
> 	/__w/dyrectorio/dyrectorio/golang/internal/grpc/grpc.go:231 +0x66a
> ```

I have found the problem. Looks like in `func RewriteUpdateErrors()` it was not handled if the err is nil. I guess the old agent got a shutdown message before it could run into that part of the code or something. I will look for a workaround on the crux's side.

m8vago:
@polaroi8d Pushed the fixes for the legacy agent update.
If you want to test it you should escape some lines in the `agent.service.ts`'s `agentVersionSupported()` function. Specifically:
```
    if (this.configService.get<string>('NODE_ENV') !== PRODUCTION) {
      return true
    }
```

Also be aware of that the agents are force pulling, so they are going to overwrite your locally built image during the updating process. :(
To fix that i've copied the GRPC_TOKEN from the new container's envs and started it with docker run.

robot9706 (approved):
🏗

golang/pkg/crane/k8s/secret.go
```diff
@@ -117,9 +115,34 @@ func (s *Secret) getSecretClient(namespace string) (v1.SecretInterface, error) {
 	return clientset.CoreV1().Secrets(namespace), nil
 }
 
-func (s *Secret) ApplyOpaqueSecret(ctx context.Context, namespace, name string, values map[string][]byte) (
+func (s *Secret) UpsertOpaqueSecrets(ctx context.Context, values map[string][]byte) (
```
nandor-magyar, line 118: Apply is a proper kubectl instruction, to avoid confusion, I would rather use ApplyWithMerge or similar name.

golang/pkg/crane/k8s/secret_store.go
```diff
@@ -0,0 +1,30 @@
+package k8s
+
+import "github.com/dyrector-io/dyrectorio/golang/internal/config"
+
+func (s *Secret) CheckPermissions() error {
+	// TODO (@m8vago, @nandor.magyar) implement
```
nandor-magyar, line 6: The http get of the getSecret will throw 403 if there is any problem with the privileges in kubernetes, I don't think it is necessary here. Wouldn't it be better off in the normal Get as an error in case of a permission problem?

nandor-magyar (approved):
Noted small things, otherwise looks good, tested in Kubernetes, it works.

web/crux-ui/locales/en/nodes.json
```diff
@@ -12,7 +12,7 @@
 
   "nodeAlreadyInUse": "This node is already in use in some deployments.",
 
-  "setup": "Install the agent",
+  "installTheAgent": "Install the agent",
```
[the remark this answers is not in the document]
m8vago: My thinking is, when i see a text on the screen and i want to locate it in the code, i just start to type it in the finder.
This one might be a bit too long, but by having the install in the json key i can easily find that.
This of course work only as long as we only have one locale.

golang/pkg/crane/k8s/secret_store.go
m8vago: This is used by the internal lib to check if we can replace a token, before we notify crux of the token replacement. It's necessary so we don't end up with the old token while crux thinks that the new one is in use.

m8vago:
> Noted small things, otherwise looks good, tested in Kubernetes, it works.

Thanks.

web/crux-ui/locales/en/nodes.json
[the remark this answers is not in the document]
nandor-magyar: I'm fine with the key, the displayed text is off.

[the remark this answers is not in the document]
m8vago: I see. Any suggestions? @polaroi8d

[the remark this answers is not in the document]
polaroi8d: Hmmm... fair point but this is not a big problem, rn IMHO.

PR #797, opened 2023-08-23
feat(web): crane metrics config

robot9706:
Added metrics (crane only) config to image and instance editors.


web/crux-ui/src/models/container.ts
```diff
@@ -180,6 +180,12 @@ export type ContainerStorage = {
   bucket?: string
 }
 
+export type Metrics = {
+  enabled?: boolean
```
polaroi8d, line 184: This shouldn't be optional, right? The default value is false.

web/crux-ui/src/models/container.ts
```diff
@@ -347,6 +355,18 @@ const mergeMarker = (image: Marker, instance: Marker): Marker => {
   }
 }
 
+const mergeMetrics = (image: Metrics, instance: Metrics): Metrics => {
+  if (!instance) {
+    return image?.enabled ? image : null
+  }
+
+  if (!image) {
```
polaroi8d, line 363: if (!image) check since it's not necessary. The first condition handle the case.

web/crux/src/app/container/container.mapper.ts
```diff
@@ -126,6 +128,18 @@ export default class ContainerMapper {
     return [...missing, ...instanceSecrets]
   }
 
+  mergeMetrics(instance: Metrics, image: Metrics): Metrics {
+    if (!instance) {
+      return image?.enabled ? image : null
+    }
+
+    if (!image) {
```
polaroi8d, line 136: Same as above.

web/crux/src/domain/container.ts
```diff
@@ -149,6 +149,12 @@ export type Storage = {
   bucket?: string
 }
 
+export type Metrics = {
+  enabled?: boolean
```
polaroi8d, line 153: Same as the default false statement.

polaroi8d (changes requested):
Works like a charm, (talked about some missing features in-person). Added minor reviews.

web/crux-ui/src/models/container.ts
robot9706: This checks if there's no instance and image metric config. The first condition only handles if there's no instance metric config.

web/crux-ui/locales/en/container.json
```diff
@@ -118,12 +118,21 @@
...
+    "ingress": "Ingress",
+    "metrics": "Metrics",
+    "metricsPath": "Metrics path",
+    "metricsPort": "Metrics port",
+    "placeholders": {
+      "headerName": "Header name",
+      "cpuUsageExample": "10m",
+      "memoryUsageExample": "512Mi",
+      "metricsPath": "/metrics",
+      "metricsPort": "80"
```
nandor-magyar, line 134: This is incorrect in the sense that it is not the default value, crane will attempt to use the first exposed port available if nothing explicit was given here.
Is this communicated any way? If not, could it be?

web/crux-ui/src/validations/container.ts
```diff
@@ -319,6 +319,16 @@ const markerRule = yup
   .nullable()
   .optional()
 
+const metricsRule = yup
```
nandor-magyar, line 322: We need an extra validation to check if the port exists in the container config, or another solution could be to change port 'text-input' to a 'select' only listing existent exposed ports.
If the port is not in the exposed ports list, service monitors will not be deployed.

web/crux-ui/locales/en/container.json
robot9706: Okay I'll rework the port selection and the placeholder.

web/crux-ui/src/models/container.ts
m8vago: Why return null when there are instance metrics?

m8vago (changes requested):

web/crux-ui/src/models/container.ts
robot9706: True, removed the extra check.

polaroi8d (approved):

PR #806, opened 2023-09-01
feat(crux-ui): useEffect submitRef update

robot9706:
- Refactor submitRef updates to use `useEffect`, this should resolve save/edit test flakyness.
- Fix an issue where unchecked registries try to fetch images on the registry detail page.


polaroi8d (approved):
LGTM (informal)

m8vago:
It would be nice to write a custom hook and use that instead.

robot9706:
@m8vago Added the submitRef to `useDyoFormik` as an optional parameter. But it might be better to have a separate hook because this way there's an extra effect which does nothing if there's no submitRef, but this seemed convenient. Thoughts?

m8vago:
Great job! There is no point to use it without a form, and we don't have forms without formik, so that's the best for it.

web/crux-ui/src/hooks/use-dyo-formik.ts
```diff
@@ -1,10 +1,25 @@
 import { FormikConfig, FormikValues, useFormik } from 'formik'
+import { MutableRefObject, useEffect } from 'react'
 
-const useDyoFormik = <Values extends FormikValues>(options: FormikConfig<Values>) =>
-  useFormik({
+const useDyoFormik = <Values extends FormikValues>(
+  options: FormikConfig<Values>,
+  submitRef?: MutableRefObject<() => Promise<any>>,
```
m8vago, line 6: It would be nice to define a custom options type for the hook, with `submitRef` included as a property, so we can provide it before the `onSubmit` definitions. That would result in cleaner usage.

m8vago (approved):

m8vago (approved):

PR #808, opened 2023-09-02
feat: add explicit port routing & go tools bump

nandor-magyar:
This contains (will contain):
- [x] go toolchain upgrades, fixes & new image
- [x] metrics port fix (use internal ports, external makes zero sense for crane)
- [x] deploymentStrategy rename and usage - it will actually be used
- [x] explicit option to select target ports when defining routing (both agent & crane)


polaroi8d (approved):
LMGT (informal)

web/crux/proto/common.proto
```diff
@@ -137,6 +137,7 @@ message Routing {
   optional string path = 101;
   optional bool stripPath = 102;
   optional string uploadLimit = 103;
+  optional uint32 port = 104;
```
m8vago, line 140: why not uint16?

m8vago (approved):

m8vago:
Tested, seems working.

web/crux/proto/common.proto
nandor-magyar: Not available in proto, or it was bugging.

PR #809, opened 2023-09-04
fix: port range exposing and various bugfixes

m8vago:
- Rework agent version check
- Fix an issue in dagent, where the port ranges weren't exposed.
- Clarify some comments
- Fix metrics port validation
- Add http -> https redirection to the main docker-compose


web/crux/src/main.ts
```diff
@@ -20,6 +20,7 @@ import PrismaErrorInterceptor from './interceptors/prisma-error-interceptor'
 import prismaBootstrap from './services/prisma.bootstrap'
 import { PRODUCTION } from './shared/const'
 import DyoWsAdapter from './websockets/dyo.ws.adapter'
+import { coerce } from 'semver'
```
robot9706, line 23: Is this used?

m8vago: Yeah, i have forget that one, the linter thrown an error as well.

polaroi8d (changes requested):
I really need some documentation regarding AGENT_SUPPORTED_MINIMUM_VERSION. We need to clarify how we intend to handle it, what it is, why we are using it, and so on. It's my fault, I didn't add this to the previous PR, sorry.

polaroi8d (changes requested):
Test case: Added a new project with 1 nginx container. Added a portrange from 7000 to 7500 same in internal and external. Added a port expose 7543 same in internal and external. Tried to deploy it and got the following error message: `client.js:1 TypeError: Cannot read properties of null (reading 'enabled')` 


Otherhand console throws a few `next-translate` error about "images" for example: 
`[next-translate] "images:" is missing in current namespace configuration. Try adding "" to the namespace "images". `

m8vago:
> add

It's not your fault, i removed this const from the last PR and added it here. I'll write some docs on it.

web/crux/src/domain/validation.ts
```diff
@@ -334,17 +335,24 @@ const createMetricsPortRule = (ports: ContainerPort[]) => {
...
+const metricsRule = yup.mixed().when(['ports'], ([ports]) => {
+  const portRule = createMetricsPortRule(ports)
+
+  return yup
     .object()
-    .shape({
-      path: yup.string().nullable(),
-      port: createMetricsPortRule(ports),
+    .when({
+      is: (it: Metrics) => it.enabled,
```
robot9706, line 344: `it` can be null so `it?.enabled` should work here.

web/crux-ui/src/validations/container.ts
```diff
@@ -331,17 +332,24 @@ const createMetricsPortRule = (ports: ContainerPort[]) => {
...
+const metricsRule = yup.mixed().when(['ports'], ([ports]) => {
+  const portRule = createMetricsPortRule(ports)
+
+  return yup
     .object()
-    .shape({
-      path: yup.string().nullable(),
-      port: createMetricsPortRule(ports),
+    .when({
+      is: (it: Metrics) => it.enabled,
```
robot9706, line 341: Same as in crux.

DEVELOPMENT.md
```diff
@@ -0,0 +1,18 @@
...
+Crux considers two things to determine if an agent is outdated:
+ - The `AGENT_PROTO_COMPATIBILITY_MINIMUM_VERSION` in the `web/crux/shared/const.ts` file
+ - And the version number in Crux's `package.json`
+
+When the agent version is between `AGENT_PROTO_COMPATIBILITY_MINIMUM_VERSION` and the `package.json`'s version
+the agent's status will be connected, and it will be fully functional. Otherwise it going to be outdated,
+and can not be used unless an update is performed.
+
+### Releasing a new agent
+While creating a new release, the `AGENT_PROTO_COMPATIBILITY_MINIMUM_VERSION` should be incremented to the version number of the new release when:
```
polaroi8d, line 16: Based on this, could you please extend the `release` target command in the root `Makefile`, please?

polaroi8d (approved):

robot9706 (approved):
Tested locally, ranges and validation seem to work fine.

PR #812, opened 2023-09-06
fix(web): validate container name duplication

m8vago:
Add client and server side validation for container name collision.


polaroi8d (approved):
Tested, works. LGTM 🪂

robot9706 (approved):
Tested locally. 🎣

PR #815, opened 2023-09-11
fix(agent): crashing on install and connection token mismatch

m8vago:
Fixes an issue, where reinstalling an agent for a different backend (issuer) caused a crash. Now we prefer the environment supplied grpc token over the saved one on issuer mismatch.

### DIFF
--- a/golang/internal/config/secret.go
+++ b/golang/internal/config/secret.go
@@ -102,27 +102,57 @@ func (c *CommonConfiguration) InjectPrivateKey(secrets SecretStore) error {
 }
 
 func (c *CommonConfiguration) InjectGrpcToken(secrets SecretStore) error {
-	token, err := secrets.GetConnectionToken()
-	if token != "" {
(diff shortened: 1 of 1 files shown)


golang/internal/config/secret.go
```diff
@@ -102,27 +102,57 @@ func (c *CommonConfiguration) InjectPrivateKey(secrets SecretStore) error {
...
+	if connectionToken == nil {
+		// there is no connection token
+		return nil
+	}
+
+	if c.JwtToken.Issuer != connectionToken.Issuer {
+		log.Info().
+			Str("environmentIssuer", c.JwtToken.Issuer).
+			Str("connectionTokenIssuer", connectionToken.Issuer).
+			Msg("Token issuer mismatch. Falling back to the environment's grpc token.")
```
polaroi8d, line 145: Minor: gRPC (sorry 🤡 )

polaroi8d (approved):
LGTM (Informal)

PR #822, opened 2023-09-14
feat(web): agent update button availability

m8vago:
- Agent update button now gets disabled, when there is no update available.
- Add `updating` as an agent status


robot9706 (approved):
Tested locally 🏗

PR #829, opened 2023-09-20
fix(crane): rename clusterrole RBAC in install script

nandor-magyar:
This removes a duplication and add the correct name for PVCs for cranes aggregated read-only cluster role.


polaroi8d (approved):
LGTM

PR #838, opened 2023-10-01
refactor: internal/helper/image_unit_test : use table driven tests

### LINKED ISSUE #832: Refactor agent image url parse helper unit tests
We have unit tests spread across various functions, but some/most of them could be organized into table driven test cases.

Refactor the tests into table driven ones where it is applicable in the file:
`golang/internal/helper/image/image_unit_test.go`

An example could be seen in `golang/internal/runtime/container/container_test.go` function `TestVersionCheck`.

PapePathe:
Fixes #832

Unit tests were spread across various functions, but most of them can be organized into table driven test cases.

### DIFF
--- a/golang/internal/helper/image/image_unit_test.go
+++ b/golang/internal/helper/image/image_unit_test.go
@@ -15,6 +15,7 @@ import (
 )
 
 type RegistryTestCase struct {
+	Name        string
 	Registry    *string
 	RegistryUrl *string
 	ExpectedUrl string
@@ -23,21 +24,25 @@ type RegistryTestCase struct {
 func TestRegistryWithTable(t *testing.T) {
 	testCases := []RegistryTestCase{
 		{
+			Name:        "Test registry url",
 			Registry:    pointer.NewPTR[string](""),
 			RegistryUrl: pointer.NewPTR[string]("test"),
 			ExpectedUrl: "test",
 		},
 		{
+			Name:        "Test registry url priority",
 			Registry:    pointer.NewPTR[string]("other"),
 			RegistryUrl: pointer.NewPTR[string]("test"),
 			ExpectedUrl: "test",
 		},
 		{
+			Name:        "Test registry url empty",
 			Registry:    nil,
 			RegistryUrl: nil,
 			ExpectedUrl: "",
 		},
 		{
+			Name:        "Test registry url registry",
 			Registry:    pointer.NewPTR[string]("other"),
 			RegistryUrl: nil,
 			ExpectedUrl: "other",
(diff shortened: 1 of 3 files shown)


golang/internal/helper/image/image_unit_test.go
```diff
@@ -13,35 +13,50 @@ import (
...
 
-func TestRegistryUrlPriority(t *testing.T) {
-	registry := "other"
-	auth := &imageHelper.RegistryAuth{
-		URL: "test",
-	}
-
-	url := imageHelper.GetRegistryURL(&registry, auth)
-	assert.Equal(t, url, "test")
+func NewPTR[T any](value T) *T {
```
nandor-magyar, line 22: This is a nice general helper function, this should be moved higher up in the hierarchy.
`golang/internal/pointer/pointer.go` it would be nice, with a simple unit test in addition.

golang/internal/helper/image/image_unit_test.go
```diff
@@ -13,35 +13,50 @@ import (
...
 )
 
-func TestRegistryUrl(t *testing.T) {
-	auth := &imageHelper.RegistryAuth{
-		URL: "test",
-	}
-
-	url := imageHelper.GetRegistryURL(nil, auth)
-	assert.Equal(t, url, "test")
+type RegistryTestCase struct {
```
nandor-magyar, line 16: The original implementation had info about what is being tested stored in the name of the function, this way now that information is lost, please add a name or description field to this struct and use it as a t.Run, if the test ever goes down, it should help identifying the problem if this test ever starts to fail somehow.

nandor-magyar (changes requested):
Hey! Thank You for your contribution. Nice and clean solution, minor things to be changed and it is good to be merged.

golang/internal/helper/image/image_unit_test.go
PapePathe: Glad you like it, i was doubtful about it. I extracted the code to `golang/internal/pointer/pointer.go`

golang/internal/helper/image/image_unit_test.go
PapePathe: Added a name field

nandor-magyar: Neat! :+1:

golang/internal/pointer/pointer.go
```diff
@@ -0,0 +1,8 @@
+package pointer
+
+/*
+ * Creates a new pointer of type T
+ */
```
nandor-magyar, line 5: Multiline comments are for package behavior description, functions generally have C++ style comments as per https://tip.golang.org/doc/comment

As simple as it is, this introduces untested code. Please add a simple unit test. This could also totally be fuzzed, what do you think?

PapePathe: Would something like the function below be sufficient.  I can learn a little about fuzz testing and add it here.
```

func TestNewPTR(t *testing.T) {
	sPtr := NewPTR[string]("")
	assert.Equal(t, *sPtr, "")

	iPtr := NewPTR[int](120)
	assert.Equal(t, *iPtr, 120)
}

```

nandor-magyar: Only if you are really interested doing so, otherwise the proposed test is fine.

PapePathe: It should be fixed now.

nandor-magyar (approved):
Looks, nice! Thank You for your efforts! :100: :+1:

nandor-magyar:
One other thing is missing here, that I should've noticed earlier, we require all commits to be signed: https://github.com/dyrector-io/dyrectorio/blob/develop/CONTRIBUTING.md#commit-signatures

PR #841, opened 2023-10-05
feat: inspect container

AmorfEvo:
- Add container inspection functionality with json and table views.
- Add better swagger descriptions and summaries on node prefix container controller endpoints.
- Move all agent timeouts to shared constants.


polaroi8d (approved):
LGTM (informal)

golang/pkg/dagent/utils/docker.go
```diff
@@ -785,3 +786,34 @@ func ContainerLog(ctx context.Context, request *agent.ContainerLogRequest) (*grp
...
+	if err != nil {
+		return "", err
+	}
+
+	prefix := request.Container.Prefix
+	name := request.Container.Name
+
+	cont, err := GetContainerByPrefixAndName(ctx, cli, prefix, name)
+	if cont == nil {
+		return "", &UnknownContainerError{}
```
nandor-magyar, line 801: I know it is not your doing, but could you replace the custom struct with a var spawned as
```
var (
ErrUnknownContainer = errors.New("... errtext")
)
```
This is cleaner looking overall, we don't need the custom struct, we don't do anything special.

nandor-magyar (approved):
Minor eye-candy only, otherwise looks good.

web/crux-ui/src/pages/[teamSlug]/nodes/[nodeId]/inspect.tsx
```diff
@@ -0,0 +1,100 @@
...
+        </div>
+
+        {inspection ? (
+          viewMode === 'table' ? (
+            <InspectTableView inspect={inspection} />
+          ) : (
+            <JsonEditor value={inspection} disabled />
+          )
+        ) : (
+          <LoadingIndicator />
```
Comment (robot9706) [inline web/crux-ui/src/pages/[teamSlug]/nodes/[nodeId]/inspect.tsx:69]: If inspection is null, the page will just load forever since no fetching happens after SSR.
So I think this is not needed.

web/crux-ui/src/models/node.ts
```diff
@@ -50,12 +50,17 @@ export type NodeDetails = DyoNode & {
   inUse: boolean
 }
 
-export const nodeConnectionOf = (node: DyoNode): NodeConnection => ({
-  address: node.address,
-  status: node.status,
-  connectedAt: node.connectedAt,
-  version: node.version,
-})
+export const nodeConnectionOf = (node: DyoNode): NodeConnection =>
```
robot9706, line 53: Node shouldn't be null, this code is only required in darklens.
You can revert this to the original code.

robot9706 (changes requested):

robot9706 (approved):

PR #848, opened 2023-10-10
fix(agent): add capital image name handling

nandor-magyar:
We found that on GitHub the organization can have uppercase letters, but the images on ghcr.io will not have uppercase, as per standards.
[Specs](https://github.com/moby/moby/blob/master/image/spec/spec.md).


m8vago (approved):

robot9706 (approved):

polaroi8d (approved):
LGTM

PR #867, opened 2023-10-21
feat(web): integrate rocket.chat

chandhuDev:
It integrates the Rocket.Chat webhook to current notification system.
And fixes the bug in the deployment by fixing the NotificationMessageTypes to align exactly with the file names in the `web/crux/assets/notificationTemplates`


web/crux/src/domain/notification-templates.ts
```diff
@@ -4,7 +4,7 @@ import { CruxInternalServerErrorException } from 'src/exception/crux-exception'
 
 const title = 'dyrector.io'
 
-export type NotificationMessageType = 'node' | 'version' | 'invite' | 'failed-deploy' | 'successful-deploy'
+export type NotificationMessageType = 'node' | 'version' | 'invite' | 'failedDeploy' | 'successfulDeploy'
```
robot9706, line 7: I think kebab-case is better here. Could you rename the template files instead?

robot9706 (changes requested):
Great work, I've tested it with a local rocket chat instance and it works fine.

There's one other thing though: You modified the database schema so please add a migration script as well, you can do it using `npx prisma migrate dev`.

chandhuDev:
can you please check now @robot9706

polaroi8d:
> can you please check now @robot9706

Linter is still failing.

PR #870, opened 2023-10-25
feat: node kick & updated agent connection kick fix & fallback token

robot9706:
- Add option to kick a connected agent
- Fix an issue with updated agents being kicked
- Agent tries to connect with the stored token first, if it fails, it falls back to the environment token if available


m8vago (approved):

web/crux-ui/locales/en/nodes.json
```diff
@@ -52,6 +52,9 @@
   "updateAvailable": "There is an update available for your node. You can update it with the update button below.",
   "updateRequired": "This agent is incompatible with the current version of dyrector.io. You can update it with the update button below.",
   "updateError": "Failed to update agent: {{error}}",
+  "kick": "Kick",
+  "areYouSureKickAgent": "Are you sure want to kick this agent?",
+  "kickingAnAgentWillStopIt": "Kicking an agent will stop it, but it won't be removed.",
```
nandor-magyar, line 57: I would add "from the host" to the end, not necessary though.

nandor-magyar (approved):

golang/internal/grpc/grpc.go
```diff
@@ -351,8 +269,118 @@ func (cl *ClientLoop) grpcLoop(connParams *ConnectionParams) {
...
+		log.Panic().Err(err).Msg("Connection refused")
+	}
+
+	if appConfig.FallbackJwtToken == nil {
+		return
+	}
+
+	log.Warn().Msg("Connection failed, trying fallback token")
+
+	err = initWithToken(grpcContext, appConfig, workerFuncs, secrets, appConfig.FallbackJwtToken)
```
nandor-magyar, line 376: You already have appConfig in params, why the `appConfig.FallbackJwtToken` again?

nandor-magyar (changes requested):

golang/internal/grpc/grpc.go
robot9706: Because `initWithToken` needs the actual token that it will use for the connection, so at first it is called with the "normal" token, then the fallback token. So eventhough I have the config, I need to know which token to use.

But if you have a better solution in mind I'm open.

nandor-magyar (approved):

PR #880, opened 2023-11-06
fix(crux-ui): sidebar vertical grow

m8vago:
Fixes sidebar scaling and vertical grow on non 16:9 displays.


robot9706 (approved):
Tested, works fine.

PR #890, opened 2023-12-06
feat(dagent): add permanent global envs and traefik upload limit

nandor-magyar:
This PR includes two small things:
First, namespace-wide shared envs was not possible it now mirrors crane functionality.
Second, Traefik upload limit is in different format compared to ingress (nginx), so if the values are provided in the latter's format it will be parsed.


golang/pkg/dagent/utils/docker.go
```diff
@@ -238,7 +239,32 @@ func DeployImage(ctx context.Context,
 	log.Debug().Str("name", deployImageRequest.ImageName).Str("full", expandedImageName).Msg("Image name parsed")
 	logDeployInfo(dog, deployImageRequest, expandedImageName, containerName)
 
-	envMap := MergeStringMapUnique(deployImageRequest.InstanceConfig.Environment, deployImageRequest.ContainerConfig.Environment)
+	if len(deployImageRequest.InstanceConfig.SharedEnvironment) > 0 {
+		err = WriteSharedEnvironmentVariables(
+			cfg.InternalMountPath,
+			getContainerPrefix(deployImageRequest),
```
m8vago, line 245: this could be lifted into a local variable.

golang/pkg/dagent/utils/environment.go
```diff
@@ -0,0 +1,85 @@
...
+import (
+	"fmt"
+	"io/fs"
+	"os"
+	"path/filepath"
+
+	"github.com/joho/godotenv"
+)
+
+const dirPerm = 0o700
```
m8vago, line 12: 600 isn't sufficient enough?

m8vago (changes requested):

golang/pkg/dagent/utils/environment.go
nandor-magyar: If it would be file, then yes, but this is for directori(es).
7 is RWX, so
R: read contents
W: write into it
X: open / cd into it
I think we need all of them.
Random note: Linux defaults to 755 for directories 644 for files.

m8vago (approved):

PR #891, opened 2023-12-07
feat: container expected state

robot9706:
Add option to specify a successful deployment condition for each image. These options are the container states shown on the deployment instance list, with an extra option to define an exit code for the `Exited` state.


m8vago (changes requested):
Let's move this to the docker config section.
Also please add a configuration option for timeout.

polaroi8d:
Please fill out the pull request description.

golang/pkg/crane/k8s/deploy_facade.go
```diff
@@ -17,6 +17,8 @@ import (
 	"k8s.io/apimachinery/pkg/api/errors"
 )
 
+const PodStateTimeoutSeconds = 120
```
nandor-magyar, line 20: It seems unused, is it?

nandor-magyar (approved):
It looks good, apart from that single unused variable.

m8vago (approved):

PR #892, opened 2023-12-15
feat(web): add encryption for third party credentials

m8vago:
Add encryption for:
- registry tokens
- storage credentials
- refactor registry and storage credential updates


web/crux/crux-docker-entrypoint.sh
```diff
@@ -5,7 +5,7 @@ case $1 in
         exec npm run start:prod
     ;;
     "migrate")
-        exec npx prisma migrate deploy
+        exec npm run encrypt:migrate
```
polaroi8d, line 8: Encrypting and migrating are two different tasks. Do we really want to combine them, it would be clearer to perform them as two separate steps, IMHO.

web/crux/package.json
```diff
@@ -21,6 +21,9 @@
     "start:dev": "nest start --watch",
     "start:debug": "nest start --debug --watch",
     "start:prod": "node dist/main",
+    "encrypt:generate": "node dist/main encrypt generate",
+    "encrypt:migrate": "node dist/main encrypt migrate",
+    "encrypt:rotate": "node dist/main encrypt rotate",
```
polaroi8d, line 26: Based on these commands encryption only works if we build the project?

web/crux/src/shared/chacha.spec.ts
```diff
@@ -0,0 +1,39 @@
+import { decryptChaCha20, encryptChaCha20, generateChacha20Key } from './chacha'
+
+describe('chachaEncryption', () => {
+  const validKey = 'aaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaa'
```
polaroi8d, line 4: Don't you want to change this string to "real" random chars?

web/crux/package.json
m8vago: Yep. Haven't checked yet if we can pass args to `nest start`.

web/crux/crux-docker-entrypoint.sh
m8vago: That's a totally fair point; however, if we already had secrets in the db they would have been lost after applying the migration.

Another possible solution would be to check on every start if we have remaining temp tables with secrets to be migrated, and process them before the startup.

Which one do you prefer?

web/crux-ui/playwright.config.ts
```diff
@@ -36,7 +36,7 @@ const config: PlaywrightTestConfig = {
     timeout: 60 * 1000, // 1 min
     reuseExistingServer: true,
   },
-  workers: CI ? 4 : undefined,
+  workers: CI ? 4 : 32,
```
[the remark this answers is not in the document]
m8vago: sure!

web/crux/package.json
polaroi8d: Okey, no probs.

PR #893, opened 2023-12-20
fix(crux-ui): hide change credentials field when creating a registry or a storage

m8vago:
Hide the change credentials toggle, while creating a registry or a storage


polaroi8d (approved):
LGTM

PR #909, opened 2024-02-14
fix(web): add missing i18n lines

polaroi8d:


web/crux-ui/locales/en/common.json
```diff
@@ -158,7 +158,9 @@
     "obsolete": "Obsolete",
     "unknown": "Unknown",
     "preparing": "Preparing",
-    "in-progress": "In progress"
+    "in-progress": "In progress",
+    "successful": "Successful",
```
m8vago, line 162: You should use `common:successful` and `common:failed` at the `t()` call instead.

polaroi8d: Updated.

web/crux-ui/src/components/projects/versions/deployments/deployment-status-tag.tsx
```diff
@@ -36,7 +36,7 @@ const statusToTextColor = (status: DeploymentStatus) => {
   }
 }
 
-const deploymentStatusTranslation = (status: DeploymentStatus): string => {
+export const deploymentStatusTranslation = (status: DeploymentStatus | 'all'): string => {
```
robot9706, line 39: What's the `all` case for?

polaroi8d: The type 'all' is not assignable to type if I'm missing from here. If there is a better approach, I'm happy to improve this. In general, I don't agree with the `deploymentStatusTranslation` function because it creates more depth complexity instead of using just the two common translations more times, but @m8vago implemented this and, as he said, I have to use it for now.

m8vago: Right now as a quick fix, I believe it's fine.
You just have add the `'all'` case to the switch statement.

m8vago (changes requested):
There are still some lines with the old translation pattern:
- line 54 in `dyo/dyrectorio/web/crux-ui/src/components/projects/versions/version-deployments-section.tsx`
- line 166 in `/home/m8/work/dyo/dyrectorio/web/crux-ui/src/components/nodes/node-deployment-list.tsx`

m8vago (approved):

PR #926, opened 2024-02-26
doc(crux): add extra info encrpytion gey generation

nandor-magyar:
The file itself is linked to the docs, so `.env.example` files have to be descriptive, since they serve as living documentation.

### DIFF
--- a/web/crux/.env.example
+++ b/web/crux/.env.example
@@ -26,6 +26,8 @@ CRUX_AGENT_ADDRESS=localhost:5000
 JWT_SECRET=jwt-secret-token
 
 # Secret key for encrypting stored credentials
+# can be generated using the CLI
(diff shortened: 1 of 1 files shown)


web/crux/.env.example
```diff
@@ -26,6 +26,8 @@ CRUX_AGENT_ADDRESS=localhost:5000
 JWT_SECRET=jwt-secret-token
 
 # Secret key for encrypting stored credentials
+# can be generated using the CLI
```
polaroi8d, line 29:
```
# Can be generated using the CLI
# Example: docker run --rm ghcr.io/dyrector-io/dyrectorio/cli/dyo:latest generate crux encryption-key
```

polaroi8d (approved):

PR #928, opened 2024-02-27
fix(web): breadcrumb navigations on list pages

m8vago:
Fix: Breadcrumb navigation on list pages while creating a new item.

### DIFF
--- a/web/crux-ui/Dockerfile
+++ b/web/crux-ui/Dockerfile
@@ -1,5 +1,5 @@
 # stage 1, build app
-FROM docker.io/library/node:19.2-alpine AS BUILDER
+FROM docker.io/library/node:20-alpine3.17 AS BUILDER
 WORKDIR /app
 
 ENV NEXT_TELEMETRY_DISABLED 1
@@ -12,7 +12,7 @@ RUN npm run build && \
     npm prune --production
 
 # stage 2, run the app
-FROM docker.io/library/node:19.2-alpine AS RUNNER
+FROM docker.io/library/node:20-alpine3.17 AS RUNNER
 WORKDIR /app
 
 ENV NODE_ENV production
--- a/web/crux-ui/Dockerfile
+++ b/web/crux-ui/Dockerfile
@@ -8,8 +8,7 @@ COPY ./package.json ./package-lock.json ./.npm ./
 RUN npm ci --arch=x64 --platform=linuxmusl --cache .npm --prefer-offline --no-fund
 
 COPY . .
-RUN npm run build && \
-    npm prune --production
+RUN npm run build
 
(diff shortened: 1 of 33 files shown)


web/crux-ui/Dockerfile
```diff
@@ -8,30 +8,31 @@ COPY ./package.json ./package-lock.json ./.npm ./
...
 EXPOSE 3000
+ENV HOSTNAME "0.0.0.0"
 ENV PORT 3000
 LABEL org.opencontainers.image.source="https://github.com/dyrector-io/dyrectorio/web/crux-ui"
 
 HEALTHCHECK --interval=30s --timeout=5s \
     CMD wget -nv -t1 --spider http://localhost:3000/ || exit 1
 
-ENTRYPOINT [ "npm", "run", "start:prod"]
+CMD [ "node", "server.js" ]
```
nandor-magyar, line 38: Using the ENTRYPOINT directive is preferable over CMD, though technically there is no much difference how it is executed.

nandor-magyar (changes requested):

web/crux-ui/Dockerfile
nandor-magyar: We had a discussion that we need the entrypoint from the original image.
```sh
#!/bin/sh
set -e

# Run command with node if the first argument contains a "-" or is not a system command. The last
# part inside the "{}" is a workaround for the following bug in ash/dash:
# https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=874264
if [ "${1#-}" != "${1}" ] || [ -z "$(command -v "${1}")" ] || { [ -f "${1}" ] && ! [ -x "${1}" ]; }; then
  set -- node "$@"
fi

exec "$@"
```

This is the content, so maybe just providing `server.js` would be enough then, to make it more apparent that there is also an entrypoint defined and it is in-use.

nandor-magyar (approved):

polaroi8d (approved):
Didn't tested in local, code looks okey for me.

web/crux-ui/e2e/utils/node-helper.ts
```diff
@@ -8,12 +8,18 @@ export const installDagent = async (page: Page) => {
   await page.goto(TEAM_ROUTES.node.list())
   await page.waitForSelector('h2:text-is("Nodes")')
 
-  await page.locator('button:has-text("Add")').click()
+  await page.waitForTimeout(1000)
+
+  page.addListener('framenavigated', frame => console.info('PW navigated', frame.url()))
```
robot9706, line 13: `console.info` :)

robot9706 (approved):

PR #942, opened 2024-03-27
fix(web): node container log stream sharing

m8vago:
- Remove NodeGlobalContainerController
- Add `_` as global container namespace.
- Refactor agent calls for log streaming
- Add upgradeable log stream watcher
- Automatically reattach to the log stream, when an agent reconnects


golang/internal/grpc/grpc.go
```diff
@@ -461,36 +462,36 @@ func executeVersionDeployRequest(
 	}
 }
 
-func streamContainerStatus(
+func streamContainerState(
```
nandor-magyar, line 465: Both Kubernetes and Docker uses Status to describe workload progress, why the change?
In general terms I would say, state is something we will alter, status is something external to ourselves that we are aware of. I'm open for discussion.

web/crux/.env.example
```diff
@@ -60,8 +60,8 @@ DISABLE_RECAPTCHA=true
 # Required only when ReCAPTCHA is enabled
 RECAPTCHA_SECRET_KEY=<recaptcha_secret_key>
 
-# Determines the quantity of lines being returned by a Docker container, while streaming its logs
-DEFAULT_CONTAINER_LOG_TAIL=100
+# Determines the maximum quantity of lines returned by a container
```
nandor-magyar, line 63: The phrasing is incorrect, the logs are not returned "by" the container. It could either be "from" or "of".

nandor-magyar (changes requested):
I feel like I have got a fresh pair of eyes :eyes: 
I've checked the Go part, that looks good. :+1: My comments are mostly wording related.
I've skimmed the TS part too, but that would better be checked by some else.

golang/internal/grpc/grpc.go
```diff
@@ -53,7 +53,7 @@ type ContainerLogContext struct {
 	Echo   bool
 }
 
-type ContainerWatchContext struct {
+type ContainerStateContext struct {
```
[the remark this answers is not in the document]
m8vago: Any suggestion?

golang/internal/grpc/grpc.go
m8vago: Docker definitely uses state:
```// ContainerState stores container's running state
// it's part of ContainerJSONBase and will return by "inspect" command
type ContainerState struct {
	Status     string // String representation of the container state. Can be one of "created", "running", "paused", "restarting", "removing", "exited", or "dead"
	Running    bool
	Paused     bool
	Restarting bool
	OOMKilled  bool
	Dead       bool
	Pid        int
	ExitCode   int
	Error      string
	StartedAt  string
	FinishedAt string
	Health     *Health `json:",omitempty"`
}```

Even though we convert that to a more simple structure,
```
export const CONTAINER_STATE_VALUES = ['running', 'waiting', 'exited', 'removed'] as const
export type ContainerState = (typeof CONTAINER_STATE_VALUES)[number]
```
we use State in the whole stack, so basically this one is the oddly named.

robot9706 (approved):

golang/internal/grpc/grpc.go
nandor-magyar: Docker does it well too. They actually manage those states, they mutate those values. If you look deeper, in that State object, you can find 'statuses', because those are external to them, since they describe external processes. So they also follow this logic, as Kubernetes. We used it vice-versa, @robot9706 started correcting them not so long ago.

golang/internal/grpc/grpc.go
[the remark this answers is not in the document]
nandor-magyar: It contains channels only, so it could have the suffix of Stream, Chan or Channels.

golang/internal/grpc/grpc.go
m8vago: So what is your proposal?

m8vago: Allrighty.

PR #944, opened 2024-04-03
feat: add error handling for agent callbacks

m8vago:
- Add a proto function for reporting agent command errors
- Report container not found errors


robot9706 (approved):
lgtm, not tested

golang/internal/grpc/grpc.go
```diff
@@ -114,6 +115,22 @@ const (
 
 var ErrConnectionRefused = errors.New("server refused connection")
 
+type AgentGrpcError struct {
+	Context    context.Context
+	InnerError error
+}
+
+func (e *AgentGrpcError) Error() string {
+	return e.InnerError.Error()
```
nandor-magyar, line 124: This can panic, the function below allows err to be nil (multiple occurrences).

golang/internal/grpc/grpc.go
```diff
@@ -567,19 +624,19 @@ func executeDeleteContainer(ctx context.Context, req *agent.ContainerDeleteReque
...
-		log.Error().Msg("Delete function not implemented")
-		return
-	}
-
-	log.Info().Msg("Deleting multiple containers")
-
-	err := deleteFn(ctx, req)
-	if err != nil {
-		log.Error().Stack().Err(err).Msg("Failed to delete multiple containers")
+func mapDeleteContainersErrorToCommandError(err *agent.AgentError) *agent.AgentCommandError {
```
nandor-magyar, line 627: These mappings could be on our "CustomError" type since each struct has its own, that  would make sense. The function then expects a type that implements an interface that has Error() and CommandError() functions. That would be a bit cleaner, what do you think?

golang/internal/common/common.go
```diff
@@ -0,0 +1,8 @@
+package common
+
+import "errors"
+
+var (
+	ErrMethodNotImplemented = errors.New("method not implemented")
+	ErrContainerNotFound    = errors.New("container not found")
```
nandor-magyar, line 7: WorkloadNotFound would be more generic, it could be pod, container or anything.

PR #945, opened 2024-04-04
fix(crux): agent token replacement status

m8vago:
Fixes an issue, where the agent status was reported as `connected` too early, while the install token was still replacing.


polaroi8d (approved):
LGTM

PR #952, opened 2024-04-10
fix(cli): use default traefik port instead url for crux_url

pavelpereyma:
This PR fix the problem of wrong value CRUX_URL when `traefikWebPort` changed by configuration.

Example:
If configuration set with `traefikWebPort=8010`, container `dyo-stable_crux-ui` will be created with env `CRUX_URL=http://dyo-stable_traefik:8010`. But actually container `dyo-stable_traefik` serve on default port 8000.


pavelpereyma:
@nandor-magyar thank you for your attention, maybe I misunderstood my problem with `dyo up`. I hope we will find the way.
I've changed in config file parameter `traefikWebPort` from 8000 to 8010. The problem is after successful `dyo up` web UI doesn't work propeply, after login I have an error "Something went wrong. Please try again later" on page and error below in "dyo-stable_crux-ui" container logs:

```
  cause: Error: connect ECONNREFUSED 172.21.0.8:8010
      at TCPConnectWrap.afterConnect [as oncomplete] (node:net:1595:16) {
    errno: -111,
    code: 'ECONNREFUSED',
    syscall: 'connect',
    address: '172.21.0.8',
    port: 8010
```

172.21.0.8 is an address of "dyo-stable_traefik" container in docker network "dyo-stable". To access to traefik in this network applications should use port that serves by traefik - 8000 (`--entrypoints.web.address=:8000`). To access to traefik via "bridge" network (outside network "dyo-stable") applications should use mapped port - 8010.
So I've changed env "CRUX_URL" from `http://dyo-stable_traefik:8010` to `http://dyo-stable_traefik:8000` for container "dyo-stable_crux-ui" and it solved my problem with web UI.

Is that right way of solving issue that I described?
And I didn't get about lock to 8080, could you please describe in more detail how my changes affect 8080?

nandor-magyar:
Ah, correct. Thank you for the clarification! 
The naming is a bit misleading then, all those constants with the prefix of `default` should indicate that those are the `internal` or `exposed` ports of containers. Renaming default to exposed or internal would be nice. You could include that in this PR if you would like to, otherwise I will record the issue.
Good job! :+1:

nandor-magyar (approved):

pavelpereyma:
Good point, I renamed `defaultTraefikWebPort` only and not sure about `defaultTraefikUIPort`, it seems fine.
And fixed my signature.

polaroi8d (approved):
Thanks @pavelpereyma for the fix. We'll merge it and release soon.

PR #964, opened 2024-04-19
fix: crane uses port/targetPort in serviceMonitors for metrics

nandor-magyar:
I used this feature this morning and it was very confusing, addressed the problem, though it changes proto, I don't think it could cause compatibility problems with crux.


polaroi8d (approved):
LGTM

PR #965, opened 2024-04-24
fix(crux-ui): add IPv6 listen for dual-stack listening

nandor-magyar:
Since **wget** defaults to IPv6 it causes healthcheck to fail, binding dualstack in crux-ui, as it is done in crux addresses the issue.


polaroi8d (approved):
LGTM

PR #966, opened 2024-04-24
feat(agent): add sysctls option to builder

nandor-magyar:
I used the sysctls options to temporarily disable IPv6 for certain containers, they could also be useful in the future.

I was receiving a deprecation notice from docker-compose since a while now, I also removed the version addressing this warning.


web/crux-ui/docker-compose.yaml
```diff
@@ -1,4 +1,3 @@
-version: '3.3'
```
polaroi8d, line 1: Why did you remove the version? There is no information in the PR about this.

polaroi8d (changes requested):

web/crux-ui/docker-compose.yaml
nandor-magyar: Because of deprecation, old versions of compose needed it, but recent version give a warning that it is a deprecated option.

polaroi8d: Okey, thanks for the clarification. (Missing from the PR description 🧒🏼)

polaroi8d (approved):

PR #970, opened 2024-05-08
chore: bump proto image version to be in sync with golang

nandor-magyar:
The protobuf generator image was left behind in version, this may cause incompatibility issues.


.github/workflows/builder_image_protobuf.yaml
```diff
@@ -37,7 +37,7 @@ jobs:
   push:
     runs-on: ubuntu-22.04
     needs: build
-    if: github.ref_name == 'develop'
+    # if: github.ref_name == 'develop'
```
polaroi8d, line 40: Why? Revert or add note to comment.

polaroi8d (changes requested):

.github/workflows/builder_image_protobuf.yaml
nandor-magyar: I'll revert before merging,  this image pipeline is far from optimal.

polaroi8d: Okey. Thanks

polaroi8d (approved):

PR #971, opened 2024-05-09
fix(crux-ui): envs split only at the first equals sign

m8vago:
- Environment values containing a `=` sign is now being parsed correctly
- Default to rwo for volumes without a type

### DIFF
--- a/web/crux-ui/src/models/compose.ts
+++ b/web/crux-ui/src/models/compose.ts
@@ -161,7 +161,7 @@ const mapVolume = (volume: string): ContainerConfigVolume => {
     id: uuid(),
     name: path ? name : '',
     path: path || name,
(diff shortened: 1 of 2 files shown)


web/crux-ui/src/models/compose.ts
```diff
@@ -161,7 +161,7 @@ const mapVolume = (volume: string): ContainerConfigVolume => {
     id: uuid(),
     name: path ? name : '',
     path: path || name,
-    type: type && CONTAINER_VOLUME_TYPE_VALUES.includes(type as VolumeType) ? (type as VolumeType) : null,
```
nandor-magyar, line 164: RWO is the reasonable default. RWX is a more special.

nandor-magyar (changes requested):

PR #978, opened 2024-05-22
feat(web): add option to disable deployment copy on version increase

m8vago:
- Add a toggle for automatic deployment copy on version increment
- Copy deployments while increasing a version depending on the said toggle.

### DIFF
--- a/web/crux-ui/locales/en/versions.json
+++ b/web/crux-ui/locales/en/versions.json
@@ -9,6 +9,7 @@
   "type": "Type",
   "incremental": "Incremental",
   "rolling": "Rolling",
+  "copyDeployments": "Copy deployments on increase",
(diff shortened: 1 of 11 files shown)


web/crux-ui/locales/en/versions.json
```diff
@@ -9,6 +9,7 @@
   "type": "Type",
   "incremental": "Incremental",
   "rolling": "Rolling",
+  "copyDeployments": "Copy deployments on increase",
```
polaroi8d, line 12: Is this correct?

polaroi8d (approved):

web/crux-ui/locales/en/versions.json
m8vago: It isn't. :)

PR #981, opened 2024-06-05
feat(web): packages

m8vago:
- Add version chains
- Add packages


web/crux-ui/locales/en/common.json
```diff
@@ -95,6 +95,7 @@
   "pipelines": "Pipelines",
   "storages": "Storages",
   "configBundles": "Config bundles",
```
polaroi8d, line 97: If you changed "config bundles" to singular, why are you leaving this as plural?

m8vago: Cause this is the translation for the side-menu, the other is the `routes.ts`.

polaroi8d: Okey, thanks for the clarification.

polaroi8d (approved):
Not tested in locally, but the code looks okey for me.

m8vago:
> Not tested in locally, but the code looks okey for me.

I don't like the current schema, gonna clarify the version relations a bit.

web/crux/src/app/version/version.service.ts
```diff
@@ -445,6 +458,10 @@ export default class VersionService {
     }
 
     const newVersion = await this.prisma.$transaction(async prisma => {
+      if (!parentVersion.chain) {
+        // we need to create the chain
```
robot9706, line 462: todo for future?

robot9706 (commented):
1 minor comment

web/crux/src/app/version/version.service.ts
m8vago: nope. forget to implement it. thanks!

PR #1000, opened 2024-09-23
feat(web): add executable to manage custom certs

nandor-magyar:
The executable to manage, load custom certs was missing. The PR addresses the issue for crux, crux-ui containers.


polaroi8d (approved):
If you tested, its okey for me.

m8vago (approved):

PR #1001, opened 2024-09-24
feat(web): add private unchecked registry

m8vago:
Add optional credentials for unchecked registries.


polaroi8d (approved):
Tested, looks okey for me.

PR #1002, opened 2024-09-26
fix(agent): add support for mixed-case image tags

nandor-magyar:
Uppercase letters in image names are not valid, however tags **can** be.


polaroi8d (approved):
LGTM.

PR #1009, opened 2024-11-06
fix(dagent): adjust of env variables, empty string as valid value

nandor-magyar:
I just ran into the issue where I wanted to override an environment variable with an empty string and it was not working properly. The change addresses the issue.


polaroi8d (approved):
LGTM

PR #1011, opened 2024-11-10
feat(crux-ui): add redirects to cards

### LINKED ISSUE #989: Feature idea: Navigate directly from Dashboard
First off, a big thank you for making dyrector.io!
While playing around with it, I noticed something that would be very handy especially for newcomers like me:
Having shortcuts from the dashboard to the relevant pages in the UI would be a nice little addition.
Currently, I can think of the following

- Users Tile -> User overview for the current team slug (/teams/{uuid})
- Projects Tile -> Project overview for the current team slug (/{slug}/projects)
- Deployments Tile + Failed Deployments Tile -> Deployments overview for the current team slug (/{slug}/deployments)
- Audit Tile -> Audit log overview for the current team slug (/{slug}/audit-log)
- Versions Tile -> Project overview for the current team slug with filter = Versioned set
  I'm unsure about this one but I figured if all other tiles get a redirect, this tile should get one too to avoid confusion 

Let me know what you think. If it's something you can imagine, I'm also open to working on this myself

curtisy1:
This PR adds links from the dashboard cards to their relating pages. Currently, the behavior is as such:

  - Users Tile -> User overview for the current team slug (/teams/{uuid})
  - Projects Tile + Versions Tile -> Project overview for the current team slug (/{slug}/projects)
  - Deployments Tile + Failed Deployments Tile -> Deployments overview for the current team slug (/{slug}/deployments)
  - Audit Tile -> Audit log overview for the current team slug (/{slug}/audit-log)

This fixes #989

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -5,7 +5,7 @@ go 1.21
 require (
 	github.com/ProtonMail/gopenpgp/v2 v2.7.1
 	github.com/go-playground/validator/v10 v10.12.0
-	github.com/golang-jwt/jwt/v4 v4.5.0
+	github.com/golang-jwt/jwt/v4 v4.5.1
 	github.com/hashicorp/go-version v1.7.0
 	github.com/ilyakaznacheev/cleanenv v1.4.2
 	github.com/pkg/errors v0.9.1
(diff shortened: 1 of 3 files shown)


polaroi8d (approved):
Looks good to me. I tested it locally, but we don't have `dashboard` scope in our conventional commits, in this case use`crux-ui` as the scope. So, the correct commit message would be `feat(crux-ui): add redirects to cards`.

curtisy1:
Fixed! Also signed off my commit to make the PR check happy

PR #1014, opened 2024-11-15
feat(cli): add host override for any number of hosts

nandor-magyar:
I wanted to use our CLI to set up everything, but not use the default hardcoded `localhost` as host header. Unfortunately, that was not possible. Here is the implementation to do so, with some usage guidelines included in the Makefile.


polaroi8d (approved):
Nice feature, LGTM.

PR #1024, opened 2024-12-04
fix: docker labels

robot9706:
Fix flaky docker label processing & add secret label type.
Implement tag filtering by creation date.
Disabled label & tag info requests for DockerHub APIs as it is rate limited and we need to fetch every manifest for a tag (TODO: rework this part).

Depends on #1018


web/crux-ui/src/validations/labels.ts
```diff
@@ -20,17 +20,21 @@ export const parseDyrectorioEnvRules = (labels: Record<string, string>): Record<
   }
 
   return Object.entries(labels).reduce((prev, [key, value]) => {
-    if (!key.startsWith('org.dyectorio.env')) {
+    if (!key.startsWith('org.dyrectorio.env')) {
```
m8vago, line 23: create a constant for `org.dyrectorio.env`

web/crux/src/domain/image.ts
```diff
@@ -44,10 +45,14 @@ export const parseDyrectorioEnvRules = (labels: Record<string, string>): Record<
...
     const params = value.split(',')
 
-    const rule: EnvironmentRule = {} as any
+    const rule: EnvironmentRule = {
+      secret: false,
+    } as any
     params.forEach(it => {
       if (it === 'required') {
         rule.required = true
+      } else if (it === 'secret') {
```
m8vago, line 54: Could not be something required and a secret in the same time?

m8vago (changes requested):

PR #1028, opened 2024-12-11
feat: advanced tag filtering

robot9706:
Implement tag filtering by creation date.


robot9706:
Merged into `fix/docker-labels` because both had to modify the V2 API client.

PR #1050, opened 2025-03-17
fix(cli): add localhost specific agent host string

nandor-magyar:
Agent could not connect using `localhost:8080` as the address.


polaroi8d (approved):
LGTM

PR #1058, opened 2025-04-10
fix: multi-db init script syntax error

polaroi8d:


polaroi8d:
Closed due the duplication.

PR #1065, opened 2025-05-08
feat: full width container config layout

m8vago:
- The container config layout now fills the whole screen instead of being arranged in 3 columns.
- Add an env variable to disable the registry image label fetching.
- Update npm packages.


polaroi8d (approved):
LGTM

PR #1071, opened 2025-05-13
feat: add env variable to disable team creation

m8vago:
- Add `DISABLE_TEAM_CREATION` to disable team creation
- Handle kratos flows with disabled registration
- Handle kratos flows with disabled password login / registration
- Link oidc account after invitation


web/crux/src/app/team/interceptors/team.creation-disabled.interceptor.ts
```diff
@@ -0,0 +1,23 @@
+import { CallHandler, ExecutionContext, Injectable, NestInterceptor } from '@nestjs/common'
+import { ConfigService } from '@nestjs/config'
+import { Observable } from 'rxjs'
+import { CruxForbiddenException } from 'src/exception/crux-exception'
+
+@Injectable()
+export default class TeamCreatinDisabledInterceptor implements NestInterceptor {
```
nandor-magyar, line 7: TeamCreatin? typo?

nandor-magyar (changes requested):
The "recovered" means that the recovery process was started, right? Not that it is already completed as it suggests.

m8vago:
> The "recovered" means that the recovery process was started, right? Not that it is already completed as it suggests.

Nope. It means the recovery step itself is completed.

nandor-magyar (approved):

PR #1090, opened 2025-09-03
fix(crane): RBAC, reorganize APIs and add missing roles

nandor-magyar:
On a fresh cluster the roles were insufficient to manage objects that crane uses.
Here is the fix.

### DIFF
--- a/web/crux/assets/install-script/install-k8s.sh.hbr
+++ b/web/crux/assets/install-script/install-k8s.sh.hbr
@@ -45,18 +45,13 @@ metadata:
   labels:
     rbac.dyrector.io/crane: "restricted-api-access"
 rules:
--
-  apiGroups:
-  - apps
-  - ""
-  - "monitoring.coreos.com"
+- apiGroups: [""]
   resources:
   - deployments
   - services
   - pods
   - ingress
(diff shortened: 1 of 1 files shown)


polaroi8d (approved):
Okey for me.

web/crux/assets/install-script/install-k8s.sh.hbr
```diff
@@ -45,18 +45,13 @@ metadata:
   labels:
     rbac.dyrector.io/crane: "restricted-api-access"
 rules:
--
-  apiGroups:
-  - apps
-  - ""
-  - "monitoring.coreos.com"
+- apiGroups: [""]
```
m8vago, line 48: is this duplicated?

m8vago: i can see now it is an object part of a list, but does the "" have meaning?

nandor-magyar: Where is it duplicated? I could not find the exact same line anywhere else.
It refers the kubernetes-core API group.

m8vago (approved):

PR #1092, opened 2025-09-12
fix(web): copying bundles during version increment

m8vago:
- Now the bundles gets copied during version increment
- Deployments of incremental versions are redeployable.


polaroi8d (approved):
Looks okey for me, if tested we can merge it.

PR #1093, opened 2025-09-12
feat: nginx proxy headers security fine-tuning

nandor-magyar:
This way CORS remains optional.


golang/pkg/crane/k8s/deploy_facade.go
```diff
@@ -213,26 +212,34 @@ func (d *DeployFacade) Deploy() error {
...
+		if len(d.params.ContainerConfig.ProxyHeaders) > 0 {
+			if err := d.configmap.deployIngressProxyHeaders(d.namespace.name, d.params.ContainerConfig.Container, d.params.ContainerConfig.ProxyHeaders...); err != nil {
+				log.Error().Err(err).Stack().Msg("Error with ingress proxy headers configmap")
+			}
+		}
+
 		if err := d.ingress.deployIngress(
 			&DeployIngressOptions{
 				namespace:     d.namespace.name,
+				buffering:     d.params.ContainerConfig.ProxyBuffering,
```
polaroi8d: Use `proxyBuffering` would be better, IMHO.

golang/pkg/crane/k8s/ingress.go
```diff
@@ -37,11 +36,12 @@ type DeployIngressOptions struct {
 	ingressPath   string
 	uploadLimit   string
 	namespace     string
-	customHeaders []string
+	proxyHeaders  []string
+	corsHeaders   []string
 	portList      []int32
 	port          uint16
+	buffering     bool
```
polaroi8d: same

polaroi8d (approved):
Add minor review.

golang/pkg/crane/k8s/deploy_facade.go
nandor-magyar: The context of Ingress already implies an underlying proxy, that is what an ingress essentially must be: a proxy. That's why I went with `buffering`. Does this help?

polaroi8d: Than do not mix it, buffering or proxybuffering?

nandor-magyar: ~~It's not mixed, it's rather two related, but separate contexts of the code.
If we are container config world it has to include some extra nouns to denote which deployment sub-component it belongs to. We could've used better typing like so, it would be much less repetitive.~~
```
containerConfig {
  ...
  ingress {
    headers
    buffering
    ...etc
  }
}
```

~~Stuttering - explicitly spelling things out multiple times is something that I would like to always avoid as it is unnecessary load for whoever reads it.
In container config it's necessary, in ingress it's implicit. If you think it's confusing for some other reasons than the name is not the exact same, let me know.~~

:upside_down_face: you can see an evolution happening.

I'm okay with changing it to `proxy-buffering` and `proxy-set-headers` as it is the actual name in nginx ingress.

polaroi8d: Its just a minor, up to you. Your decision 😊

nandor-magyar:
Tested, found a minor bug, fixed. LGTM!

PR #1097, opened 2025-10-23
feat(crux-ui): various qol changes

m8vago:
- Fix container log and deployment log autoscrolling
- Add order column to the instance list
- Add a filter option to show all containers of a node (defaults to false for docker, true for kube)
- Order projects by name


web/crux-ui/src/const.ts
```diff
@@ -46,3 +46,7 @@ export const COOKIE_TEAM_SLUG = 'crux_team_slug'
 export const TOAST_DURATION = 5000 // sec in millis
 
 export const DYO_ENV_LABEL_PREFIX = 'org.dyrectorio.env'
+export const DYO_LABEL_CONTAINER_PREFIX = 'org.dyrectorio.container.prefix'
```
polaroi8d, line 49: `DYO_LABEL_CONTAINER_PREFIX` is not consistent with the envs above. I suggest to use `DYO_LABEL_ENV_PREFIX` instead of DYO_ENV_LABEL_PREFIX` - What do you think?

web/crux-ui/src/components/nodes/use-node-details-state.tsx
```diff
@@ -88,7 +112,12 @@ const useNodeDetailsState = (options: NodeDetailsStateOptions): [NodeDetailsStat
         it.imageTag,
         utcDateToLocale(it.createdAt),
       ]),
+      containerShowAllFilter,
     ],
+    initialFilter: {
+      text: '',
+      showAll: node.type === 'k8s',
```
polaroi8d, line 119: Why added k8s type in the initial filter?

m8vago: k8s containers are not labeled yet, so we show all containers for kube nodes.

polaroi8d (approved):

polaroi8d (approved):

PR #1099, opened 2025-10-26
fix: update selector in e2e/without-login test

biocodeit:
The Playwright test in e2e/without-login was failing due to an outdated selector.
Updated the selector to match the current UI, and the test now passes ✅
Let me know if any additional changes are needed.

### DIFF
--- a/web/crux-ui/e2e/without-login/login.spec.ts
+++ b/web/crux-ui/e2e/without-login/login.spec.ts
@@ -46,7 +46,7 @@ test('sign up navigates to the register page', async ({ page }) => {
 test('recovery navigates to the recovery page', async ({ page }) => {
   await page.goto(ROUTE_LOGIN)
 
-  await page.locator('"Recovery"').click()
+  await page.locator('"Forgot your password?"').click()
 
   await expect(page).toHaveURL('/auth/recovery')
   await expect(page.locator('h1')).toContainText('Account recovery')


biocodeit:
I checked the trace on github error, there are two cases happening in Local install vs Git CI

**Local Install**
!Note the forgot password is - _Forgot your password?_
image

**Git CI**
!Note the forgot password is - _Recovery_
image

Which one to be expected.
---
Steps followed for local Install:
```
git clone git repo
sudo make up
```

polaroi8d:
Hi, whao thanks for the report, IMHO the local is the up to date, we should check the Github CI about this.

biocodeit:
okay lemme know!! I checked one previous PR where it was changed form selector `forgot your password?` to `recovery`. I am not sure why.

m8vago:
I think the difference is in the versions of dyrector.io.

It is possible that a current codebase would install the stable version, which does not include a change where the `Forgot your password?` link was changed to `Recovery`.

Obviously running the new tests on the old version will fail.
To develop e2e tests I suggest to check out the `develop` branch with git, and run the cli in the `golang/cmd/dyo` directory with:
`go run . --image-tag latest --prefix dyo-latest --host localhost up`

That that will set up an environment with the ui and the server running with the most recent development version of dyrector.io.

biocodeit:
Yes. After you said that i went into `develop` branch and the test passed. But am stuck with below error.

image

Is this know error.

PR #1113, opened 2025-12-11
feat(web): show updated tags in the instances list

m8vago:
- Show the updated tags in the deployment instances list
- Fetch tag dates for docker hub registries


web/crux-ui/src/components/composer/use-composer-state.ts
```diff
@@ -106,8 +106,8 @@ const applyEnvironments = (
     const envFile: string[] = !service.env_file
       ? []
       : typeof service.env_file === 'string'
-      ? [service.env_file]
-      : service.env_file
+        ? [service.env_file]
```
polaroi8d, line 109: These identity changes needs?

polaroi8d (approved):

web/crux-ui/src/components/composer/use-composer-state.ts
m8vago: eslint was updated, so the format rules are matching the backend's now.

PR #1118, opened 2026-01-16
chore(crane): upgrade to 1.32 API version

nandor-magyar:
With all recent changes client APIs should be up to date.

Todo: testing.


nandor-magyar:
Testing is also done, @m8vago check it please.

m8vago (approved):

PR #1134, opened 2026-04-15
feat(crane): add gateway implementation

nandor-magyar:
Adds gateway API support with custom routes.
Also includes two small fixes / adjustments to vault/secrets.


golang/pkg/dagent/utils/docker.go
```diff
@@ -488,7 +488,7 @@ func setNetwork(deployImageRequest *v1.DeployImageRequest) (networkMode string,
 func saveSecretsToVault(ctx context.Context, dog *dogger.DeploymentLogger,
 	prefix, name string, secrets map[string]string, vaultCfg *internalConfig.Vault,
 ) error {
-	if !bool(vaultCfg.BinaryAvailable) {
+	if !bool(vaultCfg.BinaryAvailable) && vaultCfg.CollectionID != "" && vaultCfg.OrgID != "" {
```
m8vago: These seem strange.

As we discussed, we should log a warning to the user, when the envs aren't null, but the binary is unavailable.

m8vago (changes requested):

m8vago (approved):

PR #1137, opened 2026-05-05
refactor(agent): vault temp folder, timeout, retry

nandor-magyar:
Added:
- timeouts
- retry logic for container status, if connection breaks
- default no timeout on Traefik


web/crux/src/app/agent/agent.service.ts
```diff
@@ -210,21 +211,60 @@ export default class AgentService {
...
+                    subject: deployment.notification.projectName,
+                    version: deployment.notification.versionName,
+                    node: deployment.notification.nodeName,
+                    owner: deployment.notification.actor,
+                  } as DeployMessage,
+                })
+                this.logger.warn(`Deployment reconnect timed out, marking as failed: ${deploymentId}`)
+              }
+            },
+            5 * 60 * 1000,
```
m8vago: please, make it a constant.

web/crux/.env.example
```diff
@@ -39,7 +39,7 @@ ENCRYPTION_SECRET_KEY=
 # The Docker image tag in the node install script
 # Uncomment to use a different agent version
 # Defaults to the version of dyrector.io
-# CRUX_AGENT_IMAGE=latest
+# CRUX_IMAGE_TAG=latest
```
m8vago: That's not the same thing. `CRUX_AGENT_IMAGE` is used as which tagged container should the installer script use for agents.

m8vago (changes requested):

web/crux/.env.example
nandor-magyar: Nice catch, very much overlooked it.

m8vago (approved):

m8vago (approved):

